PR #13, opened 2023-10-31
Migration config and db_volume ignore

TangoYankee:
- Configure mikro-orm for migrations
- Tell Eslint and tsconfig to ignore db_volume

### DIFF
--- a/src/mikro-orm.config.ts
+++ b/src/mikro-orm.config.ts
@@ -1,0 +1,19 @@
+import { TsMorphMetadataProvider } from "@mikro-orm/reflection";
+import { Migrator } from "@mikro-orm/migrations";
+import { defineConfig } from "@mikro-orm/postgresql";
+
+export default defineConfig({
+  dbName: process.env.DATABASE_NAME,
+  type: "postgresql",
+  user: process.env.DATABASE_USER,
+  password: process.env.DATABASE_PASSWORD,
+  host: process.env.DATABASE_HOST,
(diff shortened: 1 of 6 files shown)


src/mikro-orm.config.ts
```diff
@@ -0,0 +1,19 @@
+import { TsMorphMetadataProvider } from "@mikro-orm/reflection";
+import { Migrator } from "@mikro-orm/migrations";
+import { defineConfig } from "@mikro-orm/postgresql";
+
+export default defineConfig({
```
TylerMatteo, line 5: If you need this file, maybe we should remove the code where I [configured MikroOrm](https://github.com/NYCPlanning/ae-zoning-api/blob/main/src/app.module.ts#L24) in `app.module.ts`? Did you need to add duplicate configuration here to get migrations to work?

TangoYankee: I overlooked that config and forged forward thinking I would need a separate file. I'll try without the separate file and revisit.

TangoYankee: Okay, I tried to run migrations with `mikro-orm.config.ts` removed and they failed. However, I think we should still keep the configuration in `app.module.ts.` My working model says the `mikro-orm.config` is referenced for the migrations. The `app.module` is referenced by the API.

If we want to try consolidating the configurations, I think we should wait until an api ticket- when we can test both the application and the migration functionality.

TylerMatteo (approved):

Issue #15, opened 2023-10-31
Serving OpenAPI documentation


TangoYankee:
## Acceptance Criteria
- [x] Generate html with Redoc
- [x] Serve Static on Next endpoint with base path of root.

PR #17, opened 2023-10-31
Add boroughs endpoint openapi

TylerMatteo:
This PR adds an [OpenAPI](https://www.openapis.org/) compliant API description file. In addition to basic info about the API itself, it documents a `/boroughs` endpoint and the schema for the data returned by that endpoint. For the schema's properties, I assigned example values, bounds, and which are required. Note that this configuration specifies that all three properties are both _required_ (meaning they cannot be undefined) and _non-nullable_ (meaning they will always return a value of their specified type and never a literal `null`).

We don't have any tooling set up for this repo to generate and serve docs based on this OpenAPI file ([_yet_](https://github.com/NYCPlanning/ae-zoning-api/issues/15)) but I previewed it in VSCode using [this extension](https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi) and it does render successfully:
image

### DIFF
--- a/openapi.yaml
+++ b/openapi.yaml
@@ -5,12 +5,16 @@ info:
     An API for serving data related to the zoning of New York City.
   version: 0.1.0
 servers:
-  - url: https://zoning-api.planningdigital.com
+  - url: https://zoning.planningdigital.com/api
+tags:
+  - name: Boroughs
 paths:
   /boroughs:
-    get:
-      summary: Returns a list of boroughs.
+    get: 
+      summary: List boroughs
       operationId: getBoroughs
+      tags:
+      - Boroughs
       responses:
         '200':
           description: An array of borough objects.
@@ -40,3 +44,7 @@ components:
           minLength: 2
           maxLength: 2
           example: MN
+      required:
+        - id
+        - title
+        - abbr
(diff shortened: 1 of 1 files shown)


openapi.yaml
```diff
@@ -0,0 +1,50 @@
...
+components:
+  schemas:
+    Borough:
+      type: object
+      properties:
+        id:
+          type: string
+          description: A single character numeric string containing the common number used to refer to the borough. Possible values are 1-5.
+          pattern: \b[1-9]\b
+          example: 2
```
TangoYankee, line 36: This does not technically matter but will drive me crazy if I don't mention it: Manhattan is 1 (Bronx is 2).
```suggestion
          example: 1
```

TangoYankee (approved):

openapi.yaml
```diff
@@ -0,0 +1,50 @@
...
+        title:
+          type: string
+          description: The full name of the borough.
+          example: Manhattan
+        abbr:
+          type: string
+          description: The two character abbrevation for the borough.
```
TangoYankee, line 43: Spelling
```suggestion
          description: The two character abbreviation for the borough.
```

PR #48, opened 2023-11-17
46/drizzle install

### LINKED ISSUE #46: Integrate drizzle as a feature-flagged option
# Description
As a developer, I would like an opportunity to compare Mikro-orm and Drizzle orm. Mikro orm is already integrated into the application but Drizzle is not. I would like to install drizzle orm. Because applications should only be served by a single orm at a time, the use of drizzle should be kept behind a feature flag. Future tickets will implement the full application functionality. However, this ticket should also have an MVP endpoint, to verify a successful installation. An example of integrating drizzle with nest is available at [@tangoyankee/purple-hawk](https://github.com/tangoyankee/purple-hawk)

## Acceptance criteria
- [x] A feature flag is configured with an environment variable
  - [x] The variable name is 'FEATURE_FLAG_DRIZZLE' and the value is 1 
     - [x] The guard merely checks that the feature flag was set
  - [x] When the feature is disabled (the default), the application uses mikro features and configurations only
  - [x] When the feature is enabled, the application uses drizzle features and configurations only
     - [x] Endpoints that are only implemented with mikro should error when drizzle is the active orm
  - it is not necessary to change the existing npm migration scripts
- [x] There is a configuration for Drizzle migrations
- [x] There is a configuration for the application make to database requests using Drizzle
- [x] The borough endpoint is configured to serve data using Drizzle orm, as a test of the configurations
  - [x] A `schema` folder is created within the `src` directory to store drizzle schemas
  - [x] A borough schema is written to implement the borough structure in drizzle. 
    - [x] It is contained in a `borough.ts` file within the `schema` folder
  - [x] A migration is created to apply the borough schema to a fresh database
  - [x] The borough service makes its database call using drizzle orm

TangoYankee:
# Description
Configure drizzle and use it for the borough endpoint

## Deviations
The ticket states that the configuration should be dynamic and mutually exclusive between the two orms. However, disabling the configuration for Mikro caused errors importing its related files. There isn't really a simple way to also dynamically turn off these file imports based on the feature flag. So, the configuration are always on for both orms. The difference is which orm is called in the service.

The function type returns were also difficult. Each ORM represents the types of its schemas with different variables. Explicitly using these types would require an or `|` typescript operator in every instance. This would quickly get tedious. It is also not strictly necessary, as typescript infers the return type for the service functions. Consequently, I completely removed the explicit return type declaration.

#  Tickets
- closes #46


src/schema/borough.ts
```diff
@@ -0,0 +1,10 @@
+import { InferSelectModel } from "drizzle-orm";
+import { char, pgTable, text } from "drizzle-orm/pg-core";
+
+export const borough = pgTable("borough", {
+  id: char("id", { length: 1 }).primaryKey(),
+  title: text("title").notNull(),
+  abbr: text("abbr").notNull(),
+});
+
+export type SelectBorough = InferSelectModel<typeof borough>;
```
TylerMatteo, line 10: Any objection to just calling this type `Borough`? The drizzle docs [seem to use that pattern too](https://orm.drizzle.team/docs/sql-schema-declaration), though I also vaguely remember seeing this pattern in there docs somewhere.

src/schema/borough.ts
TylerMatteo, line 1: Maybe you could move this file into `/src/borough` and rename it to something like `borough.schema.ts`? If your plan is to just put schemas in this folder for now and figure out co-locating the files if/when we fully switch to Drizzle, that's cool too.

TylerMatteo (changes requested):

TylerMatteo (approved):
@TangoYankee  I'm submitting this second review as `approved` because the two things I mentioned in my previous review aren't worth holding up merging over. Overall this looks fine to me for our purposes. If we end up moving forward with Drizzle, we should eventually consider whether or not we want to implement our own "repository" classes. It's probably not necessary when you're just calling Drizzle's `.query` API (because that API is essentially already a repository) but I think calling the query builder directly in services could get unwieldy at scale.

src/schema/borough.ts
TangoYankee: I considered co-locating the files. However, my experience is that non-trivial queries requiring joining multiple schemas. Having these schemas in a common location makes the imports simpler. Going forward, it also makes the definition of the query-builder easier. On [line 23 of the db provider](https://github.com/NYCPlanning/ae-zoning-api/pull/48/files#diff-1c1f7bea169dabb11718ccd8bc9515dfe068bc4733e6bfbcb221fd0fdd87a73cR23), we pass in the schemas to the db connection. The import is simplified by keeping the schemas together in a single module.

src/schema/borough.ts
TangoYankee: Calling it 'Borough' makes sense. I had only seen the 'Select[Schema]' and 'Insert[Schema]' pattern and it was rubbing me the wrong way. The '[Schema]' and `[NewSchema]` convention strikes me as more concise and more inline with JS conventions.

TylerMatteo (approved):

PR #114, opened 2023-12-19
106 Refactor Tax Lot Endpoint - move db access from service to repo

### LINKED ISSUE #106: Refactor: Tax Lots - Move database access from service to repository layer (drizzle)
# Description
From discussion #63, our placement of database calls in the `service` files violates the separation of concerns between services and repositories. Services should handle the business logic and repositories should handle the database access.

We want to move the database access from the service to a repository. This is a drizzle-specific implementation. As such, we will need a drizzle specific repository. Mikro-orm already has the `repository` namespace: ie) tax-lot.repository.ts with a class named `TaxLotRepository`. Consequently, drizzle will use the abbreviated `repo` namespace to avoid a conflict ie) tax-lot.repo.ts with a class named `TaxLotRepo`.

## Acceptance criteria
- tax-lots
  - [x] There is a file created within the `tax-lot` folder called `tax-lot.repo.ts`
  - [x] The  `tax-lot.repo.ts` file exports a class called `TaxLotRepo`
  - [x] The `TaxLotRepo` class has `db` property with access to the drizzle `DB` provider ([tax lot example](https://github.com/NYCPlanning/ae-zoning-api/blob/5d76bbc0c59a96df36cd831fa2c092aa1e93eae8/src/tax-lot/tax-lot.repo.ts#L19))
  - [x] The `#checkTaxLotByBbl` [prepared query statement](https://orm.drizzle.team/docs/rqb/#prepared-statements) is moved to the TaxLotRepo. 
    - The execution of the prepared statement should accessed via method on the TaxLotRepo. (In line with the next listed acceptance criteria)
  - [x] Code segments within the tax lot service that directly access the database are moved to methods within the TaxLotRepo class
    - The try-catches wrapping the database access should also be moved to the method inside the TaxLotRepo
  - [x] The tax lot service calls the newly created TaxLotRepo methods
  - [x] Add `TaxLotRepo` to list of providers in the tax lot module

horatiorosa:
## Description
Moved db access from service to drizzle specific repository for the tax lot endpoint.

## Tickets
Closes #106

### DIFF
--- a/src/tax-lot/tax-lot.repo.ts
+++ b/src/tax-lot/tax-lot.repo.ts
@@ -1,12 +1,8 @@
 import { Inject } from "@nestjs/common";
 import { DB, DbType } from "src/global/providers/db.provider";
 import { eq, sql } from "drizzle-orm";
-import { DataRetrievalException, ResourceNotFoundException } from "src/error";
-import {
-  SelectTaxLot,
-  SelectTaxLotNested,
-  SelectTaxLotSpatial,
-} from "src/schema/tax-lot";
+import { DataRetrievalException } from "src/exception";
+import { SelectTaxLotSpatial } from "src/schema/tax-lot";
 import {
   taxLot,
   zoningDistrict,
@@ -30,10 +26,17 @@ export class TaxLotRepo {
     })
     .prepare("checkTaxLotByBbl");
 
-  async taxLotByBbl(bbl: string) {
-    let result: SelectTaxLotNested | undefined;
+  async checkTaxLotByBbl(bbl: string) {
     try {
-      result = await this.db.query.taxLot.findFirst({
+      return await this.#checkTaxLotByBbl.execute({ bbl });
+    } catch {
+      throw new DataRetrievalException();
+    }
+  }
+
+  async findByBbl(bbl: string) {
+    try {
+      return await this.db.query.taxLot.findFirst({
         columns: {
           boroughId: false,
           landUseId: false,
@@ -47,14 +50,11 @@ export class TaxLotRepo {
         },
       });
     } catch {
-      throw DataRetrievalException;
+      throw new DataRetrievalException();
     }
-
-    if (result === undefined) throw ResourceNotFoundException;
-    return result;
   }
 
-  async taxLotByBblGeoJson(bbl: string) {
+  async findByBblGeoJson(bbl: string) {
     let result: SelectTaxLotSpatial | undefined;
     try {
       result = await this.db.query.taxLot.findFirst({
@@ -76,36 +76,29 @@ export class TaxLotRepo {
         },
       });
     } catch {
-      throw DataRetrievalException;
+      throw new DataRetrievalException();
     }
 
-    if (result === undefined) throw ResourceNotFoundException;
+    if (result !== undefined) {
+      const geometry = JSON.parse(result.geometry) as MultiPolygon;
 
-    const geometry = JSON.parse(result.geometry) as MultiPolygon;
-
-    return {
-      type: "Feature",
-      id: result.bbl,
-      properties: {
-        bbl: result.bbl,
-        borough: result.borough,
-        block: result.block,
-        lot: result.lot,
-        address: result.address,
-        landUse: result.landUse,
-      },
-      geometry,
-    };
+      return {
+        type: "Feature",
+        id: result.bbl,
+        properties: {
+          bbl: result.bbl,
+          borough: result.borough,
+          block: result.block,
+          lot: result.lot,
+          address: result.address,
+          landUse: result.landUse,
+        },
+        geometry,
+      };
+    }
   }
 
-  async zoningDistrictByTaxLotBbl(bbl: string) {
-    let taxLotCheck: Pick<SelectTaxLot, "bbl"> | undefined;
-    try {
-      taxLotCheck = await this.#checkTaxLotByBbl.execute({ bbl });
-    } catch {
-      throw DataRetrievalException;
-    }
-    if (taxLotCheck === undefined) throw ResourceNotFoundException;
+  async findZoningDistrictByBbl(bbl: string) {
     try {
       const zoningDistricts = await this.db
         .select({
@@ -123,18 +116,11 @@ export class TaxLotRepo {
         zoningDistricts,
       };
     } catch {
-      throw DataRetrievalException;
+      throw new DataRetrievalException();
     }
   }
 
-  async zoningDistrictClassByTaxLotBbl(bbl: string) {
-    let taxLotCheck: Pick<SelectTaxLot, "bbl"> | undefined;
-    try {
-      taxLotCheck = await this.#checkTaxLotByBbl.execute({ bbl });
-    } catch {
-      throw DataRetrievalException;
-    }
-    if (taxLotCheck === undefined) throw ResourceNotFoundException;
+  async findZoningDistrictClassByBbl(bbl: string) {
     try {
       const zoningDistrictClasses = await this.db
         .select({
@@ -169,7 +155,7 @@ export class TaxLotRepo {
         zoningDistrictClasses,
       };
     } catch {
-      throw DataRetrievalException;
+      throw new DataRetrievalException();
     }
   }
 }
--- a/src/tax-lot/tax-lot.repo.ts
+++ b/src/tax-lot/tax-lot.repo.ts
@@ -1,0 +1,175 @@
+import { Inject } from "@nestjs/common";
+import { DB, DbType } from "src/global/providers/db.provider";
+import { eq, sql } from "drizzle-orm";
+import { DataRetrievalException, ResourceNotFoundException } from "src/error";
+import {
+  SelectTaxLot,
+  SelectTaxLotNested,
+  SelectTaxLotSpatial,
+} from "src/schema/tax-lot";
+import {
+  taxLot,
+  zoningDistrict,
+  zoningDistrictClass,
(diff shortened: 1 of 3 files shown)


src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+          borough: true,
+          landUse: true,
+        },
+      });
+    } catch {
+      throw DataRetrievalException;
+    }
+
+    if (result === undefined) throw ResourceNotFoundException;
+    return result;
```
TangoYankee, line 54: We'll want to keep this section in the service. It is concerned with the business logic of inspecting the data quality and deciding how to handle it.

src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+        },
+        where: (taxLot, { eq }) => eq(taxLot.bbl, bbl),
+        with: {
+          borough: true,
+          landUse: true,
+        },
+      });
+    } catch {
+      throw DataRetrievalException;
+    }
```
TangoYankee, line 51: The suggesting below (moving the undefined check to the service) will also focus this function. In turn, we can simplify its return. 

*Code segment is only for illustrative purposes. Please verify functionality in full coding environment*
```suggestion
    try {
      return await this.db.query.taxLot.findFirst({
        columns: {
          boroughId: false,
          landUseId: false,
          wgs84: false,
          liFt: false,
        },
        where: (taxLot, { eq }) => eq(taxLot.bbl, bbl),
        with: {
          borough: true,
          landUse: true,
        },
      });
    } catch {
      throw DataRetrievalException;
    }
```

src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+
+    if (result === undefined) throw ResourceNotFoundException;
```
TangoYankee, line 82: Same as above, should be kept in service

src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+    return result;
+  }
+
+  async taxLotByBblGeoJson(bbl: string) {
+    let result: SelectTaxLotSpatial | undefined;
+    try {
```
TangoYankee, line 59: Same as above, may be simplified after moving undefined check to service

src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+        )
+        .where(eq(taxLot.bbl, bbl));
+
+      return {
+        zoningDistricts,
+      };
+    } catch {
+      throw DataRetrievalException;
+    }
+  }
```
TangoYankee, line 128: We'll actually want to split this into two different methods. The service will call the methods as it needs them. 

The first method will be a public "execution" of the private prepared statement

*Code segments are only for illustrative purposes. Please verify functionality in full coding environment*
```typescript
async checkTaxLotByBbl(bbl: string) {
 try {
      return this.#checkTaxLotByBbl.execute({ bbl });
    } catch {
      throw DataRetrievalException;
    }
}
```    
The second method can keep the same name. But, it will only be concerned about actual query for zoningDistricts

```typescript
async zoningDistrictByTaxLotBbl(bbl: string) {
   try {
      return await this.db
        .select({
          id: zoningDistrict.id,
          label: zoningDistrict.label,
        })
        .from(zoningDistrict)
        .leftJoin(
          taxLot,
          sql`ST_Intersects(${taxLot.liFt}, ${zoningDistrict.liFt})`,
        )
        .where(eq(taxLot.bbl, bbl));
    } catch {
      throw DataRetrievalException;
    }
  }
}
```

The service will then control the flow of calling the repositories
```typescript
async findZoningDistrictByTaxLotBbl(bbl: string) {
  const taxLotCheck = await this.taxLotRepo.checkTaxLotByBbl(bbl);
  if (taxLotCheck=== undefined) throw ResourceNotFoundException;
  return await this.taxLotRepo.zoningDistrictClassByTaxLotBbl(bbl)
}
```

src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  async zoningDistrictClassByTaxLotBbl(bbl: string) {
```
TangoYankee, line 130: Same feedback as above regarding use of separate public `checkTaxLotByBbl` method and use of service to control flow.

TangoYankee (changes requested):
I wasn't exhaustive in pointing out all the places the feedback applies. However, it's the same general feedback for all of the functions

src/tax-lot/tax-lot.repo.ts
```diff
@@ -0,0 +1,175 @@
...
+        block: result.block,
+        lot: result.lot,
+        address: result.address,
+        landUse: result.landUse,
+      },
+      geometry,
+    };
+  }
+
+  async zoningDistrictByTaxLotBbl(bbl: string) {
```
TylerMatteo, line 101: Some feedback about method naming that I would love @TangoYankee's two cents on as well. I'm leaving this comment here as an example but this applies generally to what we name methods in services as well as repositories:

* I think we should have a convention of having some kind of verb in the name. So `zoningDistrictByTaxLotBbl` would be something like `findZoningDistrictByTaxLotBbl`. 
* I think we should avoid including words in method and variable names that are redundant in the context of where the method/variable is defined. In the case of services and repositories, this usually means taking out words that are already in the class name of the class that the method belongs too. For example, if you look at where this method is being called, we do `this.taxLotRepo.zoningDistrictByTaxLotBbl(bbl);`. A good way to keep our method names from being too huge would be to simplify this to `zoningDistrictByBbl`, so you would do `this.taxLotRepo.zoningDistrictByBbl(bbl);`. This is still a good name because `taxLotRepo` tells you that you're dealing with taxLots.
* One other minor nit pick - this should probably be `zoningDistricts` not `zoningDistrict`, because it gives back a list of districts.

Putting these suggestions together, this method name might be something like `findZoningDistrictsByBbl`.

TangoYankee: These conventions make sense to me. I also like that the class names remain singular in these conventions. In my mind, classes map to database tables, which are singular. 

In keeping with the general ticket workflow, only the method names that were created in this ticket should be updated as part of this feedback. This means there are several pre-existing classes with unexamined method names that should be scrutinized and refactored. I'll create follow-up tickets to refactor the methods on other classes.

horatiorosa: Makes sense @TylerMatteo . I will update the names. Regarding the nitpick. Can a BBL have more than one zoning district? This particular method  `findZoningDistrictByTaxLotBbl` only returns a singular zoning district.

TangoYankee: > This particular method findZoningDistrictByTaxLotBbl only returns a singular zoning district.

That is the intuitive understanding- each tax lot is associated with only one zoning district. However, zoning districts are not required to follow tax lot boundaries. This leads to situations where multiple zoning districts may intersect a tax lot (or a tax lot may not be inside any zoning district!)

horatiorosa: Makes sense, I'll update the names. 

Regarding the nitpick, can a BBL be in more than one zoning district? AFAIK, this particular method `findZoningDistrictClassByBbl` only returns a singular zoning district.

TangoYankee: > can a BBL be in more than one zoning district?

Yes, a bbl may have multiple zoning districts that intersect it. It may also be helpful to consider the difference between 'being in' and 'intersecting'. It may be natural to think of 'being in' as 'the whole tax lot is completely within the zoning district.' However, we are asking whether '**parts** of the tax lot overlap with a zoning district'. 

Because of the way zoning districts are drawn, one part of a tax lot may be in zoning district 'A' while another part of that tax lot is in zoning district 'B'. Check out `1014400022` [locally](http://localhost:3000/api/tax-lots/1014400022/zoning-districts) or on [Zola](https://zola.planning.nyc.gov/l/lot/1/1440/22#18.85/40.763761/-73.9589043) . This bbl is in two zoning districts - a `C1-9` and an `R8B`

src/tax-lot/tax-lot.service.ts
```diff
@@ -78,77 +39,15 @@ export class TaxLotService {
...
-        },
-        geometry,
-      };
+      return await this.taxLotRepo.taxLotByBblGeoJson(bbl);
     } else {
       return this.taxLotRepository.findOne({ bbl });
     }
   }
 
   async findZoningDistrictByTaxLotBbl(bbl: string) {
```
horatiorosa, line 48: @TangoYankee  The function name is also singular, shall I pluralize this as well?

src/tax-lot/tax-lot.service.ts
```diff
@@ -158,49 +57,7 @@ export class TaxLotService {
 
   async findZoningDistrictClassByTaxLotBbl(bbl: string) {
```
horatiorosa, line 58: @TangoYankee  This too.

src/tax-lot/tax-lot.service.ts
TangoYankee: We'll refactor this in a follow-up ticket.
1) It's pre-existing code and is technically out of scope of the acceptance criteria
2) We'll likely want to make several changes to this function (and several others) but we're not quite sure yet what they'll be. Better to get it in one go in the follow-up ticket.

PR #160, opened 2024-01-19
37 postgis setup: node-based

### LINKED ISSUE #37: Make PostGIS install part of repository setup

TangoYankee:
Install postgis into postgres image
Stage postgres image with setup scripts
Configure npm commmands to create postgis extension and copy data update readme setup instructions

closes #37 

This uses the existing pgClient configuration to run migrations against the database within a node environment. Though, it reads the raw sql files directly, rather than going through drizzle. It also adds npm scripts to semi-automate the running of these scripts. If we wanted to, we could combine the `db:configure, db:migrate, and db:copy` commands into a single alias. For now, this gives us the "easy single api" experience of running node scripts without the rigidity of always running an operation once it's part of the migrations.


db/query/configure.sql
```diff
@@ -0,0 +1 @@
+CREATE EXTENSION IF NOT EXISTS postgis;
```
TylerMatteo, line 1: What do you think it would take to put this in a migration? I really feel like not including this there could create headaches for us in the future. The point of migrations is to track changes to the database that power this API and be as deterministic as possible. That should include installing extensions that subsequent migrations depend on.

I know Drizzle makes it tricky to do something retroactively like this (as would any ORM with migrations, not a dig at Drizzle) but I don't think it would be that problematic to "restart" the migrations timeline for this.

db/Dockerfile
```diff
@@ -1,9 +1,14 @@
-FROM postgis/postgis:15-3.4-alpine
+FROM postgres:16-bookworm
```
TylerMatteo, line 1: This looks like the right approach to me.

TylerMatteo (changes requested):

db/Dockerfile
TangoYankee: On reflection, I think I should downgrade this back to 15-bookworm until we're ready to make a collective effort to upgrade to 16.

db/query/configure.sql
TangoYankee: I don't feel comfortable restarting the migrations for this. We should build cultural habits of respecting the integrity of the shared environments- if we restart the migrations then any outstanding instances will also need to be wiped and rebuilt. We should reserve that option as a last resort. Otherwise, we risk normalizing it.

Also, I mentally model postgis installation as closer to "database setup" than "database migration". We already accept the database needs to be in some minimally configured state before we run migrations against it. For example, we could technically initialize the zoning database with [a create database sql command](https://www.postgresql.org/docs/current/sql-createdatabase.html). However, we place this step in the "setup" phase because it practically makes more sense. The fact we're dealing with two different postgis setup pathways right now is itself evidence that it makes sense tailor the setup process based on the environment. The migrations shouldn't need to be aware of this environment.

Finally, I don't really see postgis installation creating headaches as a separate step. We already use distinct commands to handle different application setup steps. Local application setup is at least:

```
docker compose up --build
nvm use
npm i
npm run db:migrate
npm run db:load
npm run build
npm run start
```

With postgis installation as a separate step, the process is:
```
docker compose up --build
nvm use
npm i
npm run db:configure
npm run db:migrate
npm run db:load
npm run build
npm run start
```
For production releases, these steps will be codified in GitHub actions to ensure they're applied consistently.

If the concern is for developers forgetting to run `configure` before they run `migrate`, the developers should be able to recover quickly. The migrations will fail loudly because they don't have access to the specified spatial types. They will also fail safely because they're in transactions. Also, we can chain the whole setup with a single npm script like:

```"db:setup": "npm run db:configure && npm run db:migrate && npm run db:copy"```

Comment (TylerMatteo) [reply db/query/configure.sql:1]: I think we just disagree (which is OK!) about where to draw the line of what "migrations" are responsible for, though I definitely take your point that it feels arbitrary to include `CREATE EXTENSION` calls but not `CREATE DATABASE` ones. I'm OK with moving forward with your approach.

Just to explain where I'm coming from, my mental model of migrations is more that they should try to include all of the SQL that has to run for the database to be used by the application. I'm ok with having that `load` step because that is essentially our implementation of "seeding", which is typically considered separate from migrations by most ORMs that do have a mechanism for it, but I just want to make sure we don't end up with a bunch of `db:something`s down the road and the Drizzle migrations become less and less deterministic.

One last thought - if we do implement something like that `db:setup` script, we could leverage NPM's [pre and post scripts feature](https://docs.npmjs.com/cli/v9/using-npm/scripts#pre--post-scripts) if we want. Just thought - some folks avoid using that feature because it feels hand wave-y.

File: db/Dockerfile [hunk 3]
```diff
@@ -1,9 +1,14 @@
-FROM postgis/postgis:15-3.4-alpine
+FROM postgres:15-bookworm
```
TylerMatteo, line 1: Am I correct in assuming docker and docker compose are smart enough to just figure out which architecture they should use based on the user's machine? I set a `platform` param in `compose.yaml` when I was playing around with this but not sure if it was actually necessary.

TangoYankee: From the documentation, [docker says](https://docs.docker.com/engine/reference/builder/)

> By default, the target platform of the build request is used


... I have no idea what that means. But, I did not set the `platform` a param and had no issues. My sanity check was looking for a warning in the docker desktop container page. The warning says something like "container may be slow or fail because it's not meant for this platform". I had the warning for the previous image and did not get it here.

TylerMatteo: Sorry, could have been clearer. I'll rephrase the question: Does this set up use the x86 version of the image on x86 machines and the arm version on arm machines? Sounds like it does.

TylerMatteo (approved):

TylerMatteo:
@TangoYankee I approved this but, as you mentioned in stand up, it wouldn't hurt to confirm that the set up process works for other folks before merging.

pratishta (approved):
LGTM! ty ! 🚀🌕

PR #161, opened 2024-01-22
Zoning district by ID tests

### LINKED ISSUE #150: Endpoint Testing: GetZoningDistrictById, e2e and service unit
# Description

Write e2e tests and service unit tests for the `GetZoningDistrictById` endpoint

Acceptance criteria
- e2e tests
   - [x] 200 path
   - [x] 404 path from missing test
- service unit tests

pratishta:
Closes #150

### DIFF
--- a/src/zoning-district/zoning-district.repository.schema.ts
+++ b/src/zoning-district/zoning-district.repository.schema.ts
@@ -1,0 +1,8 @@
+import { z } from "zod";
+
+export const findByUuidRepoSchema = z.object({
+  id: z.string().uuid(),
+  label: z.string(),
+});
+
+export type FindByUuidRepo = z.infer<typeof findByUuidRepoSchema>;
--- a/src/zoning-district/zoning-district.repository.schema.ts
+++ b/src/zoning-district/zoning-district.repository.schema.ts
@@ -1,8 +1,6 @@
+import { zoningDistrictEntitySchema } from "src/schema";
 import { z } from "zod";
 
-export const findByUuidRepoSchema = z.object({
-  id: z.string().uuid(),
-  label: z.string(),
-});
+export const findByUuidRepoSchema = zoningDistrictEntitySchema;
 
 export type FindByUuidRepo = z.infer<typeof findByUuidRepoSchema>;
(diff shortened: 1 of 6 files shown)


test/zoning-district/zoning-district.e2e-spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        .expect(400);
+      expect(response.body.error).toBe(HttpName.BAD_REQUEST);
+    });
+
+    it("should 404 and when finding by a missing uuid", async () => {
+      const missingUuid = "03a40e74-e5b4-4faf-a8cb-a93cf6118d6c";
+      const response = await request(app.getHttpServer())
+        .get(`/zoning-districts/${missingUuid}`)
+        .expect(404);
+      expect(response.body.message).toBe(HttpName.NOT_FOUND);
```
pratishta, line 52: `NOT_FOUND` description is in the message and not error attr of response body

src/zoning-district/zoning-district.repository.schema.ts
```diff
@@ -0,0 +1,8 @@
+import { z } from "zod";
+
+export const findByUuidRepoSchema = z.object({
+  id: z.string().uuid(),
+  label: z.string(),
+});
```
TangoYankee, line 6: The schema itself looks good but we'll need to reuse it across several endpoints. For example, there is actually a tax lot endpoint that returns zoning districts (#155). For this, we can move `z.object` portion to `schema/zoning-district` and define a `zoningDistrictEntitySchema`. We can then reuse the schema for the Repo Schemas. In this case:

**Note: suggestion if for illustrative purposes. Please validate in a full code editor**
```suggestion
export const findByUuidRepoSchema = zoningDistrictEntitySchema;
```

[Tax lot entity schema example](https://github.com/NYCPlanning/ae-zoning-api/blob/886f505433ddf9ac64987a0a9b46ae13f6d54e93/src/schema/tax-lot.ts#L36)

(As a general pattern, the spatial columns are left out of the Entity Schema. If they are needed, we can [extend](https://zod.dev/?id=extend) the base EntitySchema)

TangoYankee (changes requested):
A bit of a re-usability request

src/zoning-district/zoning-district.repository.schema.ts
pratishta: Good call out, I've updated `src/schema/zoning-district.ts` to include the barebones entity schema so it can be expanded upon with more endpoints and used by other domains.

TangoYankee (approved):

PR #162, opened 2024-01-22
146 testing getTaxLotByBbl

### LINKED ISSUE #146: Endpoint Testing:  GetTaxLotByBbl, e2e and service unit
# Description

Write e2e tests and service unit tests for the `GetTaxLotByBbl` endpoint. 

## Acceptance criteria
- e2e tests
   - [x] 200 path
     - [x] bbl incorrect length
     - [x] bbl contains a letter 
   - [x] 404 path from missing bbl
- service unit tests

## Related tickets
blocked by #147 which will create the required tax-lot repository mock and tax-lot entity schema

TangoYankee:
Testing: unit and e2e for getTaxLotByBbl endpoint and service

closes #146

### DIFF
--- a/src/tax-lot/tax-lot.service.spec.ts
+++ b/src/tax-lot/tax-lot.service.spec.ts
@@ -3,7 +3,10 @@ import { TaxLotService } from "./tax-lot.service";
 import { TaxLotRepositoryMock } from "test/tax-lot/tax-lot.repository.mock";
 import { TaxLotRepository } from "./tax-lot.repository";
 import { ResourceNotFoundException } from "src/exception";
-import { getTaxLotGeoJsonByBblQueryResponseSchema } from "src/gen";
+import {
+  getTaxLotByBblQueryResponseSchema,
+  getTaxLotGeoJsonByBblQueryResponseSchema,
+} from "src/gen";
 
 describe("TaxLotController", () => {
   let taxLotService: TaxLotService;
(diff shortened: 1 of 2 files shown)


src/tax-lot/tax-lot.service.spec.ts
```diff
@@ -21,6 +24,23 @@ describe("TaxLotController", () => {
     taxLotService = moduleRef.get<TaxLotService>(TaxLotService);
   });
 
+  describe("findTaxLotByBbl", () => {
+    it("should return a tax lot when requesting a valid bbl", async () => {
```
TylerMatteo, line 28: One thing you might consider testing in "happy path" tests is `expect`ing that the service calls methods on repos that is "responsible" for calling. For instance, you might expect that your instance of `taxLotRepository` called it's `findByBbl` method with the bbl you passed in. Jest has methods such as `toHaveBeenCalledWith()` for this sort of thing.

TylerMatteo (approved):
@TangoYankee I left one comment but it's arguably overkill. If you decide to incorporate it, ping me for a re-approval.

src/tax-lot/tax-lot.service.spec.ts
TangoYankee: I place assertions like `toHaveBeenCalled...` under the umbrella of "testing implementation details". Kent C Dodds has an [article explaining how assertions against implementation details can frustrate]() the development and refactoring process.

In the case of services, developers should be confident that they can call the service method with a parameter and it will return a data structure with a specific structure. They shouldn't need to know that it calls repository methods to do this.

Testing "results" and not "implementation details" becomes more powerful when dealing with e2e tests. Take for example a tax lot endpoint. We want to refactor the controller/service/repo methods to remove the 'TaxLot' from the method names. Any 'results' based tests will be handle this refactor without issues. We just need to update the repo mock to match the real repo's refactor. However, any tests that assert that a specific controller, service, or repo calls a specific method will fail, even though the code works as its users expect.

PR #173, opened 2024-01-24
155/get zoning districts by tax lot bbl tests

### LINKED ISSUE #155: Endpoint Testing: GetZoningDistrictsByTaxLotBbl, e2e and service unit
# Description

Write e2e tests and service unit tests for the `GetZoningDistrictsByTaxLotBbl` endpoint. 

(This endpoint is within the tax lot domain and accepts a tax lot bbl. However, it returns zoning districts. In instances where we want to mock the return of an object based on its relationship to another we can use a tuple. A tuple of length two will "link" the two objects. See branch [118/unit-test](https://github.com/NYCPlanning/ae-zoning-api/blob/118/unit-test/test/tax-lot/tax-lot.repository.mock.ts#L38) for an example. We relate a tax lot to zoning districts by placing them in a tuple.)

## Acceptance criteria
- e2e tests
   - [x] 200 path
     - [x] bbl incorrect length
     - [x] bbl contains a letter 
   - [x] 404 path from missing bbl
- service unit tests

## Related tickets
blocked by #147 which will create the required tax-lot repository mock and tax-lot entity schema
blocked by #150 which will create the required zoning district entity schema

pratishta:
Closes #155 

### Mocking tax lot and zoning districts relationship
I opted to use an object with the bbl string as a computed property name that relates to the array of zoning districts 
`
[ { '1588702147': [ [Object], [Object], [Object] ] } ]
`

It looks similar to the tuple suggestion from @TangoYankee 
`
[ [ { bbl: '1588702147' }, [ [Object], [Object], [Object] ] ] ]
`

But I think that generating a computed key with the bbl string takes care of the extra overhead while still being intuitive to interpret as a relation. 

### Seeding
I updated the tax lot mocks so that we are seeding once in `checkTaxLotByBblMocks` and having other mock functions get their bbls from there. I also changed all arrays to only have one tax lot since we don't need multiple (as of now). 

I will squash the last two commits to 1 seed change commit once approved.

### DIFF
--- a/src/tax-lot/tax-lot.repository.schema.ts
+++ b/src/tax-lot/tax-lot.repository.schema.ts
@@ -1,6 +1,16 @@
 import { z } from "zod";
-import { boroughEntitySchema, landUseEntitySchema } from "src/schema";
+import {
+  boroughEntitySchema,
+  landUseEntitySchema,
+  zoningDistrictEntitySchema,
+} from "src/schema";
 import { taxLotEntitySchema } from "src/schema/tax-lot";
+
+export const checkTaxLotByBblRepoSchema = taxLotEntitySchema.pick({
+  bbl: true,
+});
+
+export type CheckTaxLotByBblRepo = z.infer<typeof checkTaxLotByBblRepoSchema>;
 
 export const findByBblRepoSchema = taxLotEntitySchema
   .omit({ boroughId: true, landUseId: true })
@@ -47,3 +57,11 @@ export const findByBblSpatialRepoSchema = findByBblRepoSchema.extend({
 });
 
 export type FindByBblSpatialRepo = z.infer<typeof findByBblSpatialRepoSchema>;
+
+export const findZoningDistrictByTaxLotBblRepoSchema = z.array(
+  zoningDistrictEntitySchema,
+);
+
+export type FindZoningDistrictByTaxLotBblRepo = z.infer<
+  typeof findZoningDistrictByTaxLotBblRepoSchema
+>;
--- a/test/tax-lot/tax-lot.repository.mock.ts
+++ b/test/tax-lot/tax-lot.repository.mock.ts
@@ -7,33 +7,27 @@ import {
 } from "src/tax-lot/tax-lot.repository.schema";
 
 export class TaxLotRepositoryMock {
-  checkTaxLotByBblMocks = Array.from(Array(1), (_, seed) =>
-    generateMock(checkTaxLotByBblRepoSchema, { seed }),
+  numberOfMocks = 1;
+
+  checkTaxLotByBblMocks = Array.from(Array(this.numberOfMocks), (_, seed) =>
+    generateMock(checkTaxLotByBblRepoSchema, { seed: seed + 1 }),
   );
 
   async checkTaxLotByBbl(bbl: string) {
     return this.checkTaxLotByBblMocks.find((row) => row.bbl === bbl);
   }
 
-  findByBblMocks = this.checkTaxLotByBblMocks.map((checkTaxLot) => {
-    const mock = generateMock(findByBblRepoSchema);
-    return {
-      ...mock,
-      ...checkTaxLot,
-    };
-  });
+  findByBblMocks = Array.from(Array(this.numberOfMocks), (_, seed) =>
+    generateMock(findByBblRepoSchema, { seed: seed + 1 }),
+  );
 
   async findByBbl(bbl: string) {
     return this.findByBblMocks.find((row) => row.bbl === bbl);
   }
 
-  findByBblSpatialMocks = this.checkTaxLotByBblMocks.map((checkTaxLot) => {
-    const mock = generateMock(findByBblSpatialRepoSchema);
-    return {
-      ...mock,
-      ...checkTaxLot,
-    };
-  });
+  findByBblSpatialMocks = Array.from(Array(this.numberOfMocks), (_, seed) =>
+    generateMock(findByBblSpatialRepoSchema, { seed: seed + 1 }),
+  );
 
   async findByBblSpatial(bbl: string) {
     return this.findByBblSpatialMocks.find((row) => row.bbl === bbl);
--- a/test/tax-lot/tax-lot.repository.mock.ts
+++ b/test/tax-lot/tax-lot.repository.mock.ts
@@ -2,22 +2,57 @@ import { generateMock } from "@anatine/zod-mock";
 import {
   findByBblRepoSchema,
   findByBblSpatialRepoSchema,
+  findZoningDistrictByTaxLotBblRepoSchema,
+  checkTaxLotByBblRepoSchema,
 } from "src/tax-lot/tax-lot.repository.schema";
 
 export class TaxLotRepositoryMock {
-  findByBblMocks = Array.from(Array(10), () =>
-    generateMock(findByBblRepoSchema, { seed: 1 }),
+  checkTaxLotByBblMocks = Array.from(Array(1), (_, seed) =>
+    generateMock(checkTaxLotByBblRepoSchema, { seed }),
   );
+
+  async checkTaxLotByBbl(bbl: string) {
+    return this.checkTaxLotByBblMocks.find((row) => row.bbl === bbl);
+  }
+
+  findByBblMocks = this.checkTaxLotByBblMocks.map((checkTaxLot) => {
+    const mock = generateMock(findByBblRepoSchema);
+    return {
+      ...mock,
+      ...checkTaxLot,
+    };
+  });
 
   async findByBbl(bbl: string) {
     return this.findByBblMocks.find((row) => row.bbl === bbl);
   }
 
-  findByBblSpatialMocks = Array.from(Array(10), () =>
-    generateMock(findByBblSpatialRepoSchema, { seed: 1 }),
-  );
+  findByBblSpatialMocks = this.checkTaxLotByBblMocks.map((checkTaxLot) => {
+    const mock = generateMock(findByBblSpatialRepoSchema);
+    return {
+      ...mock,
+      ...checkTaxLot,
+    };
+  });
 
   async findByBblSpatial(bbl: string) {
     return this.findByBblSpatialMocks.find((row) => row.bbl === bbl);
+  }
+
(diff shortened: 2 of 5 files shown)


src/tax-lot/tax-lot.repository.schema.ts
```diff
@@ -1,7 +1,17 @@
...
+import {
+  boroughEntitySchema,
+  landUseEntitySchema,
+  zoningDistrictEntitySchema,
+} from "src/schema";
 import { taxLotEntitySchema } from "src/schema/tax-lot";
 
+export const checkTaxLotByBblRepoSchema = z.object({
+  bbl: z.string().regex(RegExp("^([0-9]{10})$")),
+});
```
TangoYankee, line 11: The shape of the schema is correct. But, we can be DRYer. The checkTaxLotByBbl method is selecting the bbl field from the tax lot table, represented by the taxLotEntitySchema. We can model this with zod's [Pick](https://zod.dev/?id=pickomit) method. This also means that if we ever need to update the bbl regex, we only need to update the taxLotEntitySchema and the "check" schema will automatically get the new regex.

```suggestion
export const checkTaxLotByBblRepoSchema = taxLotEntitySchema.pick({ bbl: true });
```

test/tax-lot/tax-lot.repository.mock.ts
```diff
@@ -2,22 +2,56 @@ import { generateMock } from "@anatine/zod-mock";
...
+    generateMock(checkTaxLotByBblRepoSchema, { seed: 1 }),
+  ];
+
+  async checkTaxLotByBbl(bbl: string) {
+    return this.checkTaxLotByBblMocks.find((row) => row.bbl === bbl);
+  }
+
+  findByBblMocks = Array.from(this.checkTaxLotByBblMocks, () =>
+    generateMock(findByBblRepoSchema, {
+      stringMap: {
```
TangoYankee, line 20: Nice! I didn't know about this feature

test/tax-lot/tax-lot.repository.mock.ts
```diff
@@ -2,22 +2,56 @@ import { generateMock } from "@anatine/zod-mock";
...
+        bbl: () => this.checkTaxLotByBblMocks[0].bbl,
+      },
+    }),
   );
```
TangoYankee, line 24: Hard coding the 0 makes this a little fragile to change - we may want to add more elements later. Fortunately, we can stop using `Array.from` and switch to `[array].map`. We needed `Array.from` because we were creating arrays from nothing. But now, we can create a new array from an existing array and get direct access to the data in the existing array.

```suggestion
   findByBblMocks = this.checkTaxLotByBblMocks.map((taxLot) =>
    generateMock(findByBblRepoSchema, { stringMap: { bbl: () => taxLot.bbl } }),
  );
```

test/tax-lot/tax-lot.repository.mock.ts
```diff
@@ -2,22 +2,56 @@ import { generateMock } from "@anatine/zod-mock";
...
 
   async findByBbl(bbl: string) {
     return this.findByBblMocks.find((row) => row.bbl === bbl);
   }
 
-  findByBblSpatialMocks = Array.from(Array(10), () =>
-    generateMock(findByBblSpatialRepoSchema, { seed: 1 }),
+  findByBblSpatialMocks = Array.from(this.checkTaxLotByBblMocks, () =>
```
TangoYankee, line 30: Same `Array.from` -> `[array].map` feedback as above.

I also noticed that we swap from `checkTaxLotByBblMocks` to `findByBblMocks` in the reference. I think it makes sense to stick with the data from `checkTaxLotByBblMocks` throughout. The code sample above will satisfy this.

test/tax-lot/tax-lot.repository.mock.ts
```diff
@@ -2,22 +2,56 @@ import { generateMock } from "@anatine/zod-mock";
...
+        bbl: () => this.findByBblMocks[0].bbl,
+      },
+    }),
   );
 
   async findByBblSpatial(bbl: string) {
     return this.findByBblSpatialMocks.find((row) => row.bbl === bbl);
   }
+
+  findZoningDistrictByTaxLotBblMocks = Array.from(
```
TangoYankee, line 42: I'm seeing this has the same functionality as using [array].map (it accesses the data from the generating array). As a preference, I think it's more idiomatic to use map. I also think it help readability to use it consistently. So, I would prefer we use `map` here, too.

But, this a preference. If you want to keep `Array.from` and refactor the other functions in such a way that they use also use it, I'm fine with that. The two big priorities are 1) getting rid of the hard-coded 0s and 2) being consistent

TangoYankee (changes requested):
1) The `TaxLotController › findTaxLotByBbl › should return a tax lot when requesting a valid bbl` test fails. (If it's passing for you, let's compare environments)

2) Read all three comments in the array `from` versus `map` saga and take them as a whole. I learn along the way and my opinion evolves

src/tax-lot/tax-lot.repository.schema.ts
pratishta: Did not know about pick, updated with your suggested changes 👍

test/tax-lot/tax-lot.repository.mock.ts
pratishta: Resolving since our discussions about potentially not wanting to use `checkTaxLotByBblMocks`

pratishta:
Updated this PR with a few changes and options that are specific to commits (I'll squash appropriately when we've decided how to move forward)

Option 1 (6158eba6f89bb8f97ec5ad055e702e60a535b5fb) - Using `checkTaxLotByBbl` mocks as the initial array generation that informs the remaining mocks' bbls via the spread operator (thank you @TangoYankee for the idea in this [draft PR](https://github.com/NYCPlanning/ae-zoning-api/pull/175)). Since the spread operator merges the initially created bbls, we don't need to seed again

Option 2 (52782bd451762f03783f1c2fc8bb671bef3fd3f0) - **_Not_** using `checkTaxLotByBbl` mocks as the initial array generation and generating mocks separately instead (suggested [here](https://github.com/NYCPlanning/ae-zoning-api/pull/173#discussion_r1467680983)). The `{ seed: seed + 1 }` [here](https://github.com/NYCPlanning/ae-zoning-api/pull/173/commits/52782bd451762f03783f1c2fc8bb671bef3fd3f0#diff-c3c00167550a1e7cc56cc231548b36a48864e66e816b226ddf845e0802b6658dR13) keeps the bbls consistent across mocks. For some reason, seeding with 0 produces inconsistent results and I haven't spent enough time looking into Faker's or zod-mock's source code to determine exactly why. 

Option 3 (382353f770c092462e3a76dbe0d5b299ae41ff94) - **_Not_** using `checkTaxLotByBbl` mocks as the initial array generation and generating mocks separately instead, **_and_** just letting them be inconsistent across unrelated mocks. Since we're ultimately testing against the schema, it's fine if the bbls are not the same. 

I think Option 1 or Option 3 are fine, but would love some feedback (including whether the changes to mocking should be in a separate PR entirely).

TangoYankee:
> Updated this PR with a few changes and options that are specific to commits (I'll squash appropriately when we've decided how to move forward)
> 
> Option 1 ([6158eba](https://github.com/NYCPlanning/ae-zoning-api/commit/6158eba6f89bb8f97ec5ad055e702e60a535b5fb)) - Using `checkTaxLotByBbl` mocks as the initial array generation that informs the remaining mocks' bbls via the spread operator (thank you @TangoYankee for the idea in this [draft PR](https://github.com/NYCPlanning/ae-zoning-api/pull/175)). Since the spread operator merges the initially created bbls, we don't need to seed again
> 
> Option 2 ([52782bd](https://github.com/NYCPlanning/ae-zoning-api/commit/52782bd451762f03783f1c2fc8bb671bef3fd3f0)) - **_Not_** using `checkTaxLotByBbl` mocks as the initial array generation and generating mocks separately instead (suggested [here](https://github.com/NYCPlanning/ae-zoning-api/pull/173#discussion_r1467680983)). The `{ seed: seed + 1 }` [here](https://github.com/NYCPlanning/ae-zoning-api/pull/173/commits/52782bd451762f03783f1c2fc8bb671bef3fd3f0#diff-c3c00167550a1e7cc56cc231548b36a48864e66e816b226ddf845e0802b6658dR13) keeps the bbls consistent across mocks. For some reason, seeding with 0 produces inconsistent results and I haven't spent enough time looking into Faker's or zod-mock's source code to determine exactly why.
> 
> Option 3 ([382353f](https://github.com/NYCPlanning/ae-zoning-api/commit/382353f770c092462e3a76dbe0d5b299ae41ff94)) - **_Not_** using `checkTaxLotByBbl` mocks as the initial array generation and generating mocks separately instead, **_and_** just letting them be inconsistent across unrelated mocks. Since we're ultimately testing against the schema, it's fine if the bbls are not the same.
> 
> I think Option 1 or Option 3 are fine, but would love some feedback (including whether the changes to mocking should be in a separate PR entirely).

Let's go with option 2. I didn't realize that 0 seeds were inconsistent. We'll need consistency to make sure when we check for "missing" ids that they're truly missing and no randomness accidentally caused a match.  Good find.

test/tax-lot/tax-lot.repository.mock.ts
```diff
@@ -2,22 +2,51 @@ import { generateMock } from "@anatine/zod-mock";
...
   findByBblRepoSchema,
   findByBblSpatialRepoSchema,
+  findZoningDistrictByTaxLotBblRepoSchema,
+  checkTaxLotByBblRepoSchema,
 } from "src/tax-lot/tax-lot.repository.schema";
 
 export class TaxLotRepositoryMock {
-  findByBblMocks = Array.from(Array(10), () =>
-    generateMock(findByBblRepoSchema, { seed: 1 }),
+  NUMBER_OF_MOCKS = 1;
```
TangoYankee, line 10: Because this is held within a class, I think camelCase would be fine- I've only seen truly global variables use SCREAMING_SNAKE_CASE. But, I'm not hung up over it.

pratishta: Updated to camel case 👍

TangoYankee (approved):

Issue #168, opened 2024-01-23
Refactor: Zoning District Class By Id Schema 


horatiorosa:
Generalize the RegEx (regular expression) in the schema to be less strict 

## Acceptance Criteria
 [] RegEx should enforce starts with a single letter, case insensitive for the path parameter (actual schemas exposed to the database should remain uppercase) and at least one number
 [] RegEx should expect at least one number
 [] Service should convert path parameter to uppercase
 [] Add service test for cases where lower case is passed in the params

PR #179, opened 2024-01-29
zoning district class id

TangoYankee:
- generalize the zoning district class id; avoid enumerating all expected classes in the pattern
- accept lowercase classes in the url path parameter, converting it to uppercase in the service

closes: #168

Requires regeneration of kubb in frontend [ae-zoning-map-poc](https://github.com/NYCPlanning/ae-zoning-map-poc)


horatiorosa (approved):
LGTM
Solve my age old nit of lowercase zoning district ids breaking the app. TYSM

TylerMatteo (approved):

PR #183, opened 2024-02-05
171 & 172 document and implement paginated tax lots

### LINKED ISSUE #171: Document endpoint to get paginated tax lots
# Description

As part of discussion #97 

Create OpenAPI documentation for an endpoint which will return a paginated list of tax lots. Pagination will utilize an offset key paradigm, which exposes the "LIMIT" and "OFFSET" sql functionality through optional query parameters of the same names. When either query parameter is omitted, the application will fall back to default values. "Limit" will also have a maximum value that will be invoked if users request a number higher than this value. The returned object will return pagination metadata.

## Acceptance criteria
- create a TaxLotBasic schema:
  - [x] Returns only the boroughId and landUseId values, rather than the nested objects used by the existing TaxLot schema.
  - [x] In other words, has properties that make it resemble the `taxLotEntityShema`. Though, it is still its own distinct definition in the OpenAPI documentation
- create a Page schema that:
  - [x] describes the page meta data for paginated endpoints
  - [x] contains an 'offset' of type 'number' that specifies which offset was used in the request.
  - [x] contains a 'limit' of type 'number' that specifies which limit was used in the request
  - [x] contains a 'total' of type 'number' that specifies how many results are in the response
    - this is different from the total number of rows that match the overall request
    - when the total is less than the limit, this indicates there are no more rows that match the overall request
  - [x] contains an 'order' key of type 'string' that specifies the criteria used to sort the results
    - defaults to the primary key of the table (future iterations will support distance-based sorting)
- create an offsetParam schema that:
  - [x] describes the offset query parameter
  - [x] is optional
  - [x] has a format of a 'number-like string' (resembles a positive integer)
  - [x] has a description that politely informs users of:
    - the use of the offset
    - the restriction that it must be a positive integer
    - the 400 that will be returned if the requested offset fails to conform to restrictions
- create a limitParam schema:
  - [x] describes the limit query parameter
  - [x] is optional
  - [x] has a format of a 'number-like string' (resembles a positive integer)
  - [x] has a description that politely informs users of:
    - the use of the limit
    - the restriction that it must be a positive integer
    - the 400 that will be returned if the request value fails to conform the above restrictions
    - the default value of 20
    - the upper limit of 100 and that requests above this value will be reduced to the maximum.
- create a `tax-lots` endpoint that:
  - [x] returns a list of tax lots  
  - [x] accepts an offsetParam
  - [x] accepts an limitParam
  - [x] returns a body with a 'taxLots' key that holds a list of TaxLotBasic schemas
  - [x] returns a body that contains the Page meta data as top-level properties. (This can be achieved by extending a the Page Schema using the [`allOf` keyword](https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/)) 
  - [x] returns a 400 for invalid request
  - [x] returns a 500 for internal errors
- rebuild the open api documention page
  - [x] `npm run redoc:build`

### LINKED ISSUE #172: Implement endpoint to get paginated tax lots
# Description

As part of discussion #97 

Implement an endpoint which will return a paginated list of tax lots,as document in #171 . Pagination will utilize an offset key paradigm, which exposes the "LIMIT" and "OFFSET" sql functionality through optional query parameters of the same names. When either query parameter is omitted, the application will fall back to default values. "Limit" will also have a maximum value that will be invoked if users request a number higher than this value. The returned object will return pagination metadata.

## Acceptance Criteria
- generate zod schemas and types 
  - [x] `npm run generate`
- create a tax lot controller that
  - [x] accepts a request for a paginated list of tax lots
  - [x] accepts a query parameter for limit and offset
  - [x] uses a zod validation pipe to ensure the limit and offset parameters are valid
  - [x] calls a service for additional processing
  - [x] returns the service result to the user
- create a tax lot service that
  - [x] accepts a limit parameters of type string, defaulting to value '20'
  - [x] accepts an offset parameter of type string, defaulting to value '0'
  - [x] uses a zod schema to coerce the parameters to positive integers
    - If parsing fails, an application specific error should be thrown
    - the error should be "filtered" into a 400 response
  - [x] Reduces any 'limit' above 100 down to 100.
  - [x] calls a repository to access the applicable data
  - [x] returns an object that complies with the schema document in the open api
 - create a tax lot repository
   - [x] accepts a request for limited results of the tax lot table
   - [x] accepts an offset of type number
   - [x] accepts a limit of type number
   - [x] makes a query whose results are in the shape of 'a list of taxLotEntitySchema`
   - [x] returns the taxLotEntitySchemas 
   - [x] if there is a data retrieval error, throw the nominal error
 - test
   - [x] service unit
   - [x] e2e  

## Related
blocked by #171

TangoYankee:
## Document an endpoint to return a paginated list of tax lots

We didn't discuss a minimum for limits. But a limit of 0 has no apparent benefit. Consequently, I set it to 1.

closes #171
    
## Write a tax-lots endpoint to return a paginated list of tax lots

**Note:** *I didn't make a variable to hold the default, max, and min values for limits. This leads to a "magic number" situation. I wasn't sure of the best place for such a semi-global variable. So, I'm open to recommendations.* 

closes #172


openapi/openapi.yaml
```diff
@@ -306,6 +326,25 @@ components:
...
       description: The ten character code compromised of a one character borough, five character block, and four character lot codes.
+    limitParam:
+      name: limit
+      required: false
+      in: query
+      schema:
+        type: string
+        pattern: '^[0-9]+$'
+        example: '100'
+      description: The maximum number of results to be returned in each response. The default limit is 20 and it cannot exceed 100.
```
TylerMatteo, line 337: Are you sure about this logic? I think it's generally best to have APIs "fail loudly" rather change passed values in ways that consumers of the API might not expect.

TylerMatteo (changes requested):

openapi/openapi.yaml
TangoYankee: I agree that on principle it's good to be loud when things fail. In cases where the query parameter isn't even a number, we do fail loudly. 

But when the user has a value that is too high or too low, we know what they want- and we can partially fulfill their request. So in this instance, we "clearly renegotiate"- outlining the behavior in the docs and including the used limit in the response.

TangoYankee: Will change to throw an error

TylerMatteo (approved):

PR #187, opened 2024-02-15
Parse parameters

### LINKED ISSUE #186: Parse route parameters
# Description

Route parameters are sent as strings. We need to automatically parse them into their documented types

## Acceptance criteria
- [x] Route parameters can be documented as their true types
- [x] The parameters can be parsed from strings into their true types 
- [x] The parse definitions can be created automatically or semi-automatically

TangoYankee:
~~**Coverting to Draft** *After giving it more thought, the implicit conversion will likely cause issues with arrays of length 1, as they will look like strings to qs. More work needs to be done to validate and alleviate this concern.*~~

**Marking as ready**  
*parsing the parameters based on their json schema configurations creates dependable results*

Implement a pipe to parse parameters from strings into their documented data structures

closes #186


src/pipes/decode-params-pipe.ts
```diff
@@ -0,0 +1,49 @@
...
+      type: "string" | "number";
+    }
+  | {
+      type: "array";
+      items: {
+        type: "string" | "number";
+      };
+    };
+
+export class DecodeParamsPipe<T extends ZodRawShape> implements PipeTransform {
```
TylerMatteo, line 16: The code here lgtm but I think it could do with some commenting so that someone looking at this code for the first time understands what it's doing and why it's necessary. You could add it at the top of the class, or inside the `transform` method, or a combination of both.

TylerMatteo (changes requested):

src/pipes/decode-params-pipe.ts
TangoYankee: Wilco

Issue #176, opened 2024-01-26
Document spatial filter for tax lots endpoint


TangoYankee:
# Description

As part of discussion #97 

Document a spatial filter for the endpoint which accesses many tax lots. This extension filters the tax lots based on a geometry passed through the query parameters. The endpoint will accept the geometry type,  coordinates, and a buffer. It will the tax lots which intersect the geometry 

## Acceptance criteria
- create a geometryParam that:
  - [x] describes the type of geometry being requested with this query param
  - [x] is optional
  - [x] is an enum with options of `Point`, `LineString`, and `Polygon`
  - [x] has a description that politely informs the user
    - the role of the parameter
    - that it must be provided when using the spatial filter
    - that it adds additional restrictions to the coordinates based on the type
      - Point type must be of length one
      - LineString must be of length between 2 and 5, inclusive
      - Polygon must be of length between 4 and 5, inclusive
- create a lonsParam that
  - [x] specifies the longitude coordinates being requested with this query param
  - [x] is optional
  - [x] is an array of numbers in the format of an ["unexploded form array"](https://swagger.io/docs/specification/serialization/#query) with length between 1 and 5, inclusive
  - [x] has a description that politely informs the user
    - the role of the parameter
    - that it must be provided when using the spatial filter
    - that it must form a valid geometry when combined with data from the other spatial parameters
- create a latsParam that
  - [x] specifies the latitude coordinates being requested with this query param
  - [x] is optional
  - [x] is an array of numbers in the format of an ["unexploded form array"](https://swagger.io/docs/specification/serialization/#query) with length between 1 and 5, inclusive
  - [x] has a description that politely informs the user
    - the role of the parameter
    - that it must be provided when using the spatial filter
    - that it must form a valid geometry when combined with data from the other spatial parameters
- create a bufferParam that:
  - [x] specifies the spatial buffer being requested with this query param
  - [x] is optional
  - [x] is a number
  - [x] has a description that politely informs the user
    - it provides a spatial buffer around the geometry
    - is in feet
    - that it is optional when using the spatial buffer
- update the tax-lots endpoint
  - [x] Add the geometryParam, lonsParam, latsParam, and bufferParam query parameters
  - [x] Add instructions to the description regarding the spatial filter
    - It is useful for finding the tax lots that intersect a geometry
    - When applying the spatial filter, results are sorted by distance from the center of the largest portion of the shape.
[...]

PR #189, opened 2024-02-20
176/document spatial filter

TangoYankee:
Write the open api documentation for the query parameters that comprise a geospatial filter for tax lots

closes https://github.com/NYCPlanning/ae-zoning-api/issues/176


openapi/openapi.yaml
```diff
@@ -70,9 +70,15 @@ paths:
       parameters: 
         - $ref: "#/components/parameters/limitParam"
         - $ref: "#/components/parameters/offsetParam"
+        - $ref: "#/components/parameters/geometryParam"
```
TylerMatteo, line 73: I wouldn't hold up the PR over this, and it's sort of out of scope anyway but I wanted to flag we might want to reconsider this `somethingParam` naming convention we've adopted. It's nitpicky but it just feels overly verbose to tack `Param` onto everything when the refs are already namespaced by `#/components/parameters`, unless there is some other name collision problem I'm missing.

TylerMatteo (approved):

PR #190, opened 2024-03-05
Implement tax lot spatial filter

### LINKED ISSUE #177: Implement spatial filter on tax lots endpoint
# Description

As part of discussion #97 

Implement the spatial filter for the endpoint which accesses many tax lots. This extension filters the tax lots based on a geometry passed through the query parameters. The endpoint will accept the geometry type,  coordinates, and a buffer. It will the tax lots which intersect the geometry

## Acceptance criteria
- generate kubb types and schemas
  - [x] ```npm run generate```
- update tax lot controller
  - [x] ensure the update zod schema rejects query parameters that do not meet the documented formats
- update the tax lot service
  - [x] validate that if one of the spatial query parameters are invoked, then all of the required spatial parameters are present
  - [x] parse the lats and lons into proper JS arrays
  - [x] ensure the lats and lons arrays are of equal length
  - [x] ensure the lengths of the lats and lons arrays are within the bounds for their geometry type
  - [x] construct a geojson object from the query parameters
  - [x] for linestrings and polygons, leverage a repository call to ensure the shape is valid
  - [x] if validation fails at any stage, throw a validation error
  - [x] leverage the repository to find a list of tax lots that meet the spatial filter credentials
- update the tax lot service
  - [x] create a method that will validate the geometry
  - [x] create a method that will find the tax lots that intersect the geometry and its buffer. 
  - [x] the spatial filter method must sort by distance to the geometry. For polygons, use the center of the maximum inscribed circle of the geometry 
- update the exception filtering
  - [x] Update the Client Request filter to throw a 400 for any geometry validation errors

TangoYankee:
Add ability to filter tax lots based query parameters with spatial information

closes #177

### DIFF
--- a/src/tax-lot/tax-lot.repository.ts
+++ b/src/tax-lot/tax-lot.repository.ts
@@ -17,7 +17,14 @@ import {
   FindZoningDistrictsByBblRepo,
   FindZoningDistrictClassesByBblRepo,
   FindManyRepo,
+  FindManyBySpatialFilterRepo,
+  FindGeomBufferRepo,
+  FindMaximumInscribedCircleCenterRepo,
+  CheckGeomIsValidRepo,
+  FindGeomFromGeoJsonRepo,
 } from "./tax-lot.repository.schema";
+import { Geometry } from "geojson";
+import { Geom } from "src/types";
 
 export class TaxLotRepository {
   constructor(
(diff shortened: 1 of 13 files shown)


src/tax-lot/tax-lot.repository.ts
```diff
@@ -70,6 +77,100 @@ export class TaxLotRepository {
     }
   }
 
+  async findGeomFromGeoJson(
```
TylerMatteo, line 80: Similar to the other commit on `checkGeomIsValid`, could the return type of this just be `Geom`?

src/tax-lot/tax-lot.repository.ts
```diff
@@ -70,6 +77,100 @@ export class TaxLotRepository {
...
+      const result = await this.db.execute(
+        sql`SELECT ST_Buffer(${geom}, ${buffer}) AS buffer`,
+      );
+      return result.rows[0] as { buffer: Geom };
+    } catch {
+      throw new DataRetrievalException();
+    }
+  }
+
+  async checkGeomIsValid(geom: Geom): Promise<CheckGeomIsValidRepo> {
```
TylerMatteo, line 108: Could the return type of this just be `boolean`? I know you're using a type infered from a zod schema but I think you could have a zod schema that is an instance of `z.boolean()` instead of an object containing an instance of `z.boolean()`.

src/tax-lot/tax-lot.repository.ts
```diff
@@ -17,7 +17,14 @@ import {
...
+  FindManyBySpatialFilterRepo,
+  FindGeomBufferRepo,
+  FindMaximumInscribedCircleCenterRepo,
+  CheckGeomIsValidRepo,
+  FindGeomFromGeoJsonRepo,
 } from "./tax-lot.repository.schema";
+import { Geometry } from "geojson";
+import { Geom } from "src/types";
 
 export class TaxLotRepository {
```
TylerMatteo, line 29: A lot of these geospatial methods strike me as things that can be reused, possibly into some sort of general purpose Repository class for geospatial entities. I'm cool with putting that off until we have more than just the Tax Lot repo that need them, but wanted to make a note of it.

src/tax-lot/tax-lot.repository.ts
```diff
@@ -70,6 +77,100 @@ export class TaxLotRepository {
...
+        sql`SELECT ST_Transform(ST_GeomFromGeoJSON(${shape}), CAST(${targetSrid} AS INTEGER)) AS geom`,
+      );
+      return result.rows[0] as { geom: Geom };
+    } catch {
+      throw new DataRetrievalException();
+    }
+  }
+
+  async findGeomBuffer(
+    geom: Geom,
```
TylerMatteo, line 95: I don't mind passing around PostGIS-encoded geoms within Repositories but I do think we should consider long term if having public methods on Repos that take in these geoms is the best approach. Having PostGIS-encoded geoms as arguments feels like leaking data access concerns out of the repository.

TylerMatteo (changes requested):
Simplifying the return types on `findGeomFromGeoJson` and `checkGeomIsValid` are the only changes I'm concerned about for getting this merged in. The other comments are just notes or things to consider in the future.

PR #195, opened 2024-03-19
expand postgres commands

damonmcc:
related to https://github.com/NYCPlanning/ae-data-flow/issues/2

## changes
- configure npm commands to use admin credentials for operations that require them
- add npm script and command to create databases in the DO cluster for developers

## background

Locally deployed databases are useful, but when the API is deployed it won't use a locally deployed database. When trying to run the commands in this repo against the (already deployed) database cluster in DO called `zoning-api`, it became clear that there are two types of connections: those that require admin privileges and those that do not.

In the issue https://github.com/NYCPlanning/ae-data-flow/issues/2, I described a potential approach to databases and users that was inspired by the one used for PFF (described [here](https://github.com/NYCPlanning/ae-private/discussions/15#discussioncomment-8764252)). These changes enable that implementation.

## screenshots

a development db created by running these commands, structured using `pg:migrate`, and modified using in-progress commands in the new [`ae-data-flow`](https://github.com/NYCPlanning/ae-data-flow) repo:

Screenshot 2024-03-18 at 10 11 46 PM


db/create-dev-database.ts
```diff
@@ -0,0 +1,26 @@
+import "dotenv/config";
+import { exit } from "process";
+import { pgClientAdminDefaultDB } from "./pg-client";
```
pratishta, line 3: How does this create a new DB with `DATABASE_NAME` when it's set to `database: "defaultdb",` in pg-client? Does the env variable `DATABASE_NAME` override it?

pratishta (dismissed):
Able to create a new DB in DO 👍

TangoYankee (changes requested):
It seems this is a good time to check in and validate the workflow. Reading through the code and the supporting tickets, I want to make sure we're not muddying the workflow between developing the application and developing the data.

damonmcc:
per chat today (notes [here](https://github.com/NYCPlanning/ae-data-flow/issues/2#issuecomment-2009831628)), will rework this to be useful for a new approach. if no changes here are needed for a new approach, will close this

damonmcc:
per relevant issue [here](https://github.com/NYCPlanning/ae-data-flow/issues/2) this PR is no longer needed because AE's data flow process will only touch the API DB at the final stage: replacing rows in tables

we're using development DBs for data flow dev and tabling the idea of development DBs for API dev

PR #197, opened 2024-03-29
Issue/191 intergrate sentry io using nestjs sentry

horatiorosa:
Spike into using nestjs-sentry
400 errors are picked up by Sentry.io


.sentryclirc
TylerMatteo, line 1: Please delete this file and add it to `.gitignore`. I'll revoke this token in Sentry just to be safe.

TylerMatteo (changes requested):

.sentryclirc
horatiorosa: completed. 
🗂️  -->  🗑️

TylerMatteo (approved):

PR #198, opened 2024-03-29
use data flow repo and allow remote Zoning API databases

damonmcc:
related to https://github.com/NYCPlanning/ae-data-flow/issues/29

## summary of changes

These changes enable the use of remote Zoning API databases and deprecate the first-pass approach for populating those DBs. The use of a remote DB seems essential for (an eventually) remote API instance.

## details

When trying to use a remote Zoning API database in Digital Ocean, all API calls resulted in a 500 error without the SSL mode configuration. An example of error details is shown below (sensitive values replaces with `*`s).
```json
{
  "statusCode": 500,
  "error": "Internal Server Error",
  "name": "DataRetrievalException",
  "message": "Error while retrieving data: no pg_hba.conf entry for host \"**.***.***.***\", user \"****\", database \"****\", no encryption",
}
```

This also deprecates the old method of populating a Zoning API DB (copying csvs from a public S3 bucket) in favor of the new method (using the `ae-data-flow` [repo](https://github.com/NYCPlanning/ae-data-flow))

## screenshots

see comment [here](https://github.com/NYCPlanning/ae-data-flow/issues/29#issuecomment-2027795955) showing successful use of new data flow repo to populate a Zoning API DB


src/global/providers/db.provider.ts
```diff
@@ -18,6 +18,9 @@ export const DbProvider: FactoryProvider = {
       user: dbConfig.user,
       password: dbConfig.password,
       database: dbConfig.name,
+      ssl: {
+        rejectUnauthorized: false,
```
TangoYankee, line 22: If memory serves, we use also this provider for production configurations and in those environments we'll want this to be true. This is likely a small change to something like ```rejectUnauthorized: process.env.NODE_ENV === "production"```

README.md
```diff
@@ -12,36 +12,36 @@ Once you have cloned this repo, install the necessary dependencies:
 npm i
 ```
 ### Set up your `.env` file
-Create a file called `.env` in the root folder of the project and copy the contents of `sample.env` into that new file
+Create a file called `.env` in the root folder of the project.
 
-### Run local database with docker compose
+If using a local database, copy the contents of `sample.env` into that new file. If using the remote database, use our password manager to find credentials for the dev database.
+
+### (Optional) Use a local database
```
TangoYankee, line 19: This phrasing is fine for a first step. But, I think it has implications on how AE manages database environments. This makes it worth spinning out into a new issue for discussion and refined documentation. See #199

TangoYankee (changes requested):
1. Need to follow up on the `rejectUnauthorized` parameter.

2. Could someone walk me through setting up a local database with the data flow?

src/global/providers/db.provider.ts
damonmcc: (per offline chat and likely a massive oversimplification of the details)

this change is needed locally because we're using http and not https

a deployed API server shouldn't require these changes because it'd have an issued certificate

damonmcc: @TangoYankee see [014aa8d](https://github.com/NYCPlanning/ae-zoning-api/pull/198/commits/014aa8d601a3e6b5d59b4cae8cf7d53e136a9a19)

PR #218, opened 2024-05-02
Add Agency, Managing Code, and Project schemas

### LINKED ISSUE #210: Add Agency, Managing Code, and Project Drizzle schemas to Zoning API
Implement the following tables as Drizzle schemas in Zoning API per @TangoYankee's ERD [seen here](https://github.com/NYCPlanning/ae-data-flow/blob/commitment-erd/sql/commitment-erd.png). You can also find a rough implementation of the tables in raw sql [here](https://github.com/NYCPlanning/ae-data-flow/blob/commitment-erd/sql/commitment.sql) for reference.

- [x] Define schemas for the `project`, `managing_code`, and `agency` tables as drizzle schemas in `/src/schema`. Note that the 
- [x] Table columns match names and data types from ERD. Note that the ERD defines snake_case columns names as they should appear in the database. The properties on the typescript schemas should have the equivalent camelCase name. (example of how to have custom name for composite key available [here](https://orm.drizzle.team/docs/indexes-constraints#composite-primary-key)) 
- [x] Table foreign/primary key relationships are defined as per the ERD. Note that the `project` table has a composite primary key made up of `id` and `managing_agency`
- [x] For now, do not implement geospatial columns - these are the ones that start with `mercator` or `li_ft` in the ERD
- [x] Define `project_category` enum and use it in `project` table schema

bmarchena:
Closes #210

### DIFF
--- a/src/schema/managing-code.ts
+++ b/src/schema/managing-code.ts
@@ -1,0 +1,12 @@
+import { char, pgTable } from "drizzle-orm/pg-core";
+import { z } from "zod";
+
+export const managingCode = pgTable("managing_code", {
+  id: char("id", { length: 3 }).primaryKey(),
+});
+
+export const managingCodeEntitySchema = z.object({
+  id: z.string().regex(RegExp("^[A-Z][0-9]{3}")),
+});
+
+export type ManagingCodeEntity = z.infer<typeof managingCodeEntitySchema>;
--- a/src/schema/agency.ts
+++ b/src/schema/agency.ts
@@ -1,0 +1,14 @@
+import { pgTable, text } from "drizzle-orm/pg-core";
+import { z } from "zod";
+
+export const agency = pgTable("agency", {
+  initials: text("initials").primaryKey(),
+  name: text("name"),
+});
+
+export const agencyEntitySchema = z.object({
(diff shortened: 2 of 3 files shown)


src/schema/managing-code.ts
```diff
@@ -0,0 +1,12 @@
+import { char, pgTable } from "drizzle-orm/pg-core";
+import { z } from "zod";
+
+export const managingCode = pgTable("managing_code", {
+  id: char("id", { length: 3 }).primaryKey(),
+});
+
+export const managingCodeEntitySchema = z.object({
+  id: z.string().regex(RegExp("^[A-Z][0-9]{3}")),
```
TangoYankee, line 9: Nit: This regex is something to keep an eye on. I'm realizing now that I never documented the expected shape of the managing codes. All of the characters are expected to be numeric and the regex could be more restrictive to reflect this: ```^[0-9]{3}```

But we should also question whether we want to use a regex at all. zod provides a .length method that could also meet our needs and be more performant than a regex: ```z.string().length(3)```

TangoYankee (approved):
Good work! My comment was just food for thought.  Let's keep things moving to the next ticket

PR #222, opened 2024-05-09
Add commitment schema

### LINKED ISSUE #214: Add commitment schema to API
# Description
Add [`commitment`](https://github.com/NYCPlanning/ae-data-flow/blob/f513c4a44751497b48a6926a3047da2b1ead9ec9/sql/commitment.sql#L197) and [`commitment_fund`](https://github.com/NYCPlanning/ae-data-flow/blob/f513c4a44751497b48a6926a3047da2b1ead9ec9/sql/commitment.sql#L209) schemas to api

## Acceptance criteria
- [x] A `commitment` schema exists and mirrors the columns documented in the ERD/reference SQL table
- [x] A `commitment_fund` schema exists and mirrors the columns documented in the ERD/reference SQL table
- [x] The database columns use snake_case and the javascript variables use camelCase

## Related tickets
Blocked by #213

bmarchena:
Closes #214

### DIFF
--- a/src/schema/commitment.ts
+++ b/src/schema/commitment.ts
@@ -1,0 +1,47 @@
+import {
+  char,
+  date,
+  foreignKey,
+  pgTable,
+  text,
+  uuid,
+} from "drizzle-orm/pg-core";
+import { z } from "zod";
+import { managingCodeEntitySchema } from "./managing-code";
+import { project } from "./project";
+import { budgetLine } from "./budget-line";
+
+export const commitment = pgTable(
+  "commitment",
+  {
+    id: uuid("id").primaryKey(),
+    type: char("type", { length: 4 }),
+    plannedDate: date("planned_date"),
+    managingCode: char("managing_code", { length: 3 }),
+    projectId: text("project_id"),
+    budgetLineCode: text("budget_line_code"),
+    budgetLineId: text("budget_line_id"),
+  },
+  (table) => {
+    return {
+      projectFk: foreignKey({
+        columns: [table.managingCode, table.projectId],
+        foreignColumns: [project.managingCode, project.id],
+      }),
+      budgetLineFk: foreignKey({
+        columns: [table.budgetLineCode, table.budgetLineId],
+        foreignColumns: [budgetLine.code, budgetLine.id],
+      }),
+    };
+  },
+);
+
+export const commitmentEntitySchema = z.object({
+  id: z.string().uuid(),
+  type: z.string().length(4),
+  plannedDate: z.date(),
+  managingCode: managingCodeEntitySchema,
+  projectId: z.string(),
+  budgetLineCode: z.string(),
+  budgetLineId: z.string(),
+});
--- a/src/schema/commitment-fund.ts
+++ b/src/schema/commitment-fund.ts
@@ -1,0 +1,22 @@
+import { numeric, pgTable, uuid } from "drizzle-orm/pg-core";
+import { projectCategoryEnum } from "./project";
+import { z } from "zod";
+import { commitment } from "./commitment";
+
+export const commitmentFund = pgTable("commitment_fund", {
+  id: uuid("id").primaryKey(),
+  commitmentId: uuid("commitment_id").references(() => commitment.id),
+  category: projectCategoryEnum("fund_category"),
+  value: numeric("value"),
+});
+
+export const commitmentFundEntitySchema = z.object({
+  id: z.string().uuid(),
+  commitmentId: z.string().uuid(),
+  category: z.string(),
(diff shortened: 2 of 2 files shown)


src/schema/project.ts
```diff
@@ -39,5 +39,7 @@ export const projectEntitySchema = z.object({
   id: z.string(),
   managingAgency: agencyEntitySchema,
   description: z.string(),
+  minDate: z.date(),
+  maxDate: z.date(),
```
bmarchena, line 43: Noticed these were missing while working on the commitment schemas

src/schema/commitment.ts
```diff
@@ -0,0 +1,57 @@
...
+    type: char("type", { length: 4 }),
+    plannedDate: date("planned_date"),
+    managingCode: char("managing_code", { length: 4 }),
+    projectId: text("project_id"),
+    budgetLineCode: text("budget_line_code"),
+    budgetLineId: text("budget_line_id"),
+  },
+  (table) => {
+    return {
+      fk: foreignKey({
```
TangoYankee, line 28: This is an interesting scenario where a single table has several foreign keys. The documentation doesn't really specify how to handle this situation and I'm not actually sure how we're expected to handle it. I see two paths ahead.  First, we could use this approach where the table column is mapped to a foreign column by its position. Alternatively, we might want to actually split the multicolumn keys into their own object values and then also move the `type` foreign key to a `references` method.

```js
  "commitment",
  {
    id: uuid("id").primaryKey(),
    type: char("type", { length: 4 }).references(() => commitmentType.code),
    plannedDate: date("planned_date"),
    managingCode: char("managing_code", { length: 4 }),
    projectId: text("project_id"),
    budgetLineCode: text("budget_line_code"),
    budgetLineId: text("budget_line_id"),
  },
  (table) => {
    return {
      managingCodeProjectId: foreignKey({
        columns: [
          table.managingCode,
          table.projectId,
        ],
        foreignColumns: [
          project.managingCode,
          project.id,
        ],
        name: "managing_code_project_id_fk",
      }),
 budgetLineCodebudgetLineId: foreignKey({
        columns: [
          table.budgetLineCode,
          table.budgetLineId,
        ],
        foreignColumns: [
          budgetLine.code,
          budgetLine.id,
        ],
        name: "budget_line_code_budget_line_id_fk",
      }),
    };
  },
```


I want to test which approaches actually work and from there choose which to use. The way we can do this is make some prototype migrations and see whether they create tables that truly match the expected schemas. I will mess around with this and get back to you.

TangoYankee (changes requested):

src/schema/commitment.ts
TangoYankee: Okay, I generated some prototype migrations with each of the two approaches.

The schema as written will generate [single multicolumn foreign key](https://github.com/NYCPlanning/ae-zoning-api/commit/f4cb1250dbdb32992442877bf768b8c381d918af#diff-fbc99d2ca2b8ee1300f5c5139329e48ed5c6813d03ebc5b77abbcd30f492e198R102). 

The schema as shown in the code snippet in the example above will generate [three distinct foreign key constraints](https://github.com/NYCPlanning/ae-zoning-api/commit/60213701cb9dd6d16b9389bd68e2b836c6db61fa#diff-9f8ba36e2ab40174bda01dff9860838c6866adaa524bb08349211884d4e03ca4R102-R114). This is the desired outcome. Let's go with the approach shown in the code snippet.

I removed the "name" option from the code sample because drizzle will generate constraint names automatically. These generated constraint names will meet the requirement of being unique within the table. They are also more descriptive because they reflect the name of the columns they're composed of.

TangoYankee:
@bmarchena The change to the zod schema in the project file drew my attention to a missing foreign key in the drizzle schema. I made ticket #223 to capture these changes. Since we're messing with git anyway, could you break the project zod schema changes into a branch to address #223? On that branch, also add the foreign key mentioned in the new ticket

src/schema/commitment.ts
```diff
@@ -0,0 +1,47 @@
...
+import { project } from "./project";
+import { budgetLine } from "./budget-line";
+
+export const commitment = pgTable(
+  "commitment",
+  {
+    id: uuid("id").primaryKey(),
+    type: char("type", { length: 4 }),
+    plannedDate: date("planned_date"),
+    managingCode: char("managing_code", { length: 4 }),
```
TangoYankee, line 20:
```suggestion
    managingCode: char("managing_code", { length: 3 }),
```

TangoYankee (changes requested):

TangoYankee (approved):

PR #260, opened 2024-05-14
Add redoc build step to lint-staged

### LINKED ISSUE #259: Build Open API static site as part of pre-commit hooks
# Description
We can ensure that updates to the openapi.yaml are turned into their new site by making the redoc:build process part of the precommit hook.

## Acceptance criteria
- [x] The redoc build process is part of the pre-commit hook

TangoYankee:
Ensures that open api site stays in step with documentation

Behavior to look for when reviewing
- If you make a change to `openapi/openapi.api`, then commiting should automatically run the `redoc:build` command and regenerate the `openapi/index.html` file.
- If you make a change somewhere else in the repository (let's say you were updating something in the source code) and the `openapi/openapi.yaml` file is unchanged, then `openapi/index.html` should be skipped

closes #259

### DIFF
--- a/package.json
+++ b/package.json
@@ -106,6 +106,7 @@
     "*.{ts,tsx}": [
       "bash -c 'npm run typecheck'",
       "npm run lint"
-    ]
+    ],
+    "openapi/openapi.yaml" : "bash -c 'npm run redoc:build && git add openapi/index.html'"
   }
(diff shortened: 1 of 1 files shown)


dhochbaum-dcp (approved):
Works as expected

PR #265, opened 2024-05-14
216/migration

### LINKED ISSUE #216: Migrate database to integrate capital project database schemas
# Description
Create a migration to integrate the capital project schemas into the database. 

## Tips
- This is a volatile process. It must only be done on a local database. This prevents anyone else from looking at the database and having their workflow disrupted. It simplifies the management of databases as they can be easily thrown away/iterated on, without the risk of orphaned databases in the cloud.
- Creating the migration files
  - We use [`drizzle-kit`](https://orm.drizzle.team/kit-docs/overview) cli tool with the [`generate:pg`](https://orm.drizzle.team/kit-docs/commands#generate-migrations) command to generate the up migrations.
  - `drizzle-kit` is installed as a dev dependency. However, we do not have an npm script which leverages `drizzle-kit` to create migrations. Instead, you'll need to use `npx` to access and run `drizzle-kit`. ex) ```npx drizzle-kit generate:pg```
  - `drizzle-kit` does not create down-migrations. If you want to reset your local database to a previous state, you'll need to 1) [drop](https://orm.drizzle.team/kit-docs/commands#drop-migration) the errant migration(s) and 2) remove your `db-volume` directory to wipe the database state. Yes, this is annoying. However, It's only really necessary when making a PR. When developing the feature, make as many "up" migrations as you need to get the schemas into the correct format. Only at the end, drop all the "incremental" migrations. Then generate a single migration that contains the correct state of the schemas. This single migration will make it to the PR.
- Running the migrations
   - `drizzle-kit` does not actually apply the migrations
   - We rely on a custom `migrate` file to apply the migrations
   - The migrations are run with a custom `drizzle:migrate` script
   - If you want more debug information when running migrations, you can edit the `db/migrate.ts` file
- Editing the migration files
  -  drizzle-kit is going to mess up the geometry types. Specifically, it's going to put quotes around the name. We'll need to manually remove the quotes from the names of the geometry files.
   - Migrations are stateful. They are tracked in a table within the database. If a migration has already been successfully applied, editing the migration file will not update the database; drizzle will mark the migration as already applied and skip it. If you want to edit an already-applied migration, you'll need to delete the `db-volume` to wipe the state of the database. From there, you can re configure the database and re-apply the migrations.

## Acceptance Criteria
- [ ] A single migration file is created to apply the capital project schemas

## Related tickets
Blocked by #215 
**UPDATE** Also blocked by #221 to include relevant geographies
**UPDATE v2** blocked by #223 and #224

bmarchena:
closes #216


TangoYankee (changes requested):
1) The `capital_project` table, `managing_agency` column needs a foreign key to the `agency` table, `initials` column
2) The `capital_project` table, `capital_project_category` column should have its name shortened to just `category`
3) The  `capital_project_fund` table, `capital_project_category` column should
    - rename the column to `capital_fund_category` (it should have been this way in the original ERD. The errors really stacked with this one)
    - change the column type to `capital_fund_category`

We're now in the phase where we have to worry about the thing I said "don't worry about"- managing migration versions during local development. I'll schedule a time to work through it together.

Issue #229, opened 2024-05-09
Open API: Create Capital Project Schema


TangoYankee:
# Description

Create a reusable `CapitalProject` schema

## Acceptance criteria
- [x] A `CapitalProject` schema exists under [`components/schemas`](https://github.com/NYCPlanning/ae-zoning-api/blob/9561c9bd8b80a8ca49f52505899bfec4500904e9/openapi/openapi.yaml#L406)
  - Is placed in alphabetical order within the list of schemas, as practicable
- [x] it is of type 'object'
- [x] it has `id` property:
   - type string
   - required
   - description: {helpful description}
   - example: `HWPEDSF5`
- [x] it has `managingCode` property:
   - type string
   - required
   - description: {helpful description}
   - pattern: regex meaning number-like with length of 3
   - example: `850`
- [x] it has a `description` property
   - type string
   - required
   - description: {helpful description}
   - example: `Multi-Site Pedestrian Safety Phase 5`
- [x] it has `managingAgencyInitials` property:
   - type string
   - required
   - description: {helpful description}
   - example: `DOT`
- [x] it has `minDate` property:
   - type date
   - required
   - description: {helpful description}
   - example: {[grab a date from executing a toJSON](https://stackoverflow.com/a/15952652/13729626)}
- [x] it has `maxDate` property:
   - type date
   - required
   - description: {helpful description}
   - example: {[grab a date from executing a toJSON](https://stackoverflow.com/a/15952652/13729626)}
- [x] it has `category` property
   - references `CapitalProjectCategory` from #230 

## Helpful example
[Borough schema](https://github.com/NYCPlanning/ae-zoning-api/blob/9561c9bd8b80a8ca49f52505899bfec4500904e9/openapi/openapi.yaml#L407)

## Related tickets

blocked by #230 

## EDIT: added description field

PR #267, opened 2024-05-15
229 create capital project schema for open api

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/229


openapi/openapi.yaml
```diff
@@ -470,7 +470,44 @@ components:
...
         - "Lump Sum"
         - "ITT, Vehicles and Equipment"
-      descirption: The type of Capital Project.
+      description: The type of Capital Project.
+    CapitalProject:
+      type: object
+      properties:
+        id:
+          type: string
+          description: A string of variable length, composite of managing code and project ID
```
TangoYankee, line 479: There is no single field that represents the whole of the composite key. The composite key exists as a combination of the "id" and "managingCode". I recognize the ambiguity of calling it 'id'; it makes it read like a thing that should be unique. Unfortunately, I don't think we're going find a perfect name. We just need to remember that `id` is not the composite key. It is one of the two fields that compromises the composite key. 

```suggestion
          description: The id for the project, which combines with the managing code to make a unique id
```

TangoYankee (changes requested):

openapi/openapi.yaml
```diff
@@ -470,7 +470,44 @@ components:
         - "Fixed Asset"
         - "Lump Sum"
         - "ITT, Vehicles and Equipment"
-      descirption: The type of Capital Project.
+      description: The type of Capital Project.
```
TangoYankee, line 473: nice catch

openapi/openapi.yaml
pratishta: Wait so the example of this is `HWPEDSF5`-- is that the ID that gets combined with a separate managing code? and `HWPEDSF` big is just part of the project ID?

pratishta: Ok yes because managing codes are numbers. Okay cool

TangoYankee (dismissed):

TangoYankee (approved):

Issue #278, opened 2024-05-17
Endpoint Tracking Meta-Ticket


TangoYankee:
# Description
This is a handy summary of seeing which endpoints are documented and implemented. 

~~We created the `stable` tag to tell the frontend which version of the openapi documentation is actually implemented. However, it depends on every endpoint being implemented in the same order that it was added to the documentation. This meta ticket tracks the order we want to implement the endpoints. Correspondingly, it tells use the order we must add the endpoint documentation to keep the `stable` tag helpful. We can still write the endpoints in parallel. But, the merge order must match the order shown below. Also, we are free to change the order before we merge the documentation. However, once we merge the documentation, we are committed to that order.~~

~~The endpoints are listed in the order of desired implementation. There is a checkbox for when it has been documented and when it has been implemented. Implementation order must follow documentation order to keep the stable tag meaningful. The stable tag will be placed at the latest point in time where the endpoint has been both documented and implemented.~~

~~As we merge documentation onto main, let's also track the commit hash where the documentation was added. That will make it easier to move the stable tag to that commit once the endpoint has been implemented.~~

## Endpoint Order
- [x] findAgencies
  - [x] documented
  - [x] implemented
- [x] findCommunityDistrictsByBoroughId
  - [x] documented
  - [x] implemented
- [x] findCityCouncilDistricts
  - [x] documented
  - [x] implemented
- [x] findCapitalProjectsByBoroughIdCommunityDistrictId
  - [x] documented 
  - [x] implemented
- [x] findCapitalProjectsByCityCouncilDistrictId
  - [x] documented
  - [x] implemented
- [x] findCapitalProjectByManagingCodeCapitalProjectId
  - [x] documented 
  - [x] implemented
- [x] findCapitalProjectTiles
  - [x] documented 
  - [x] implemented
- [x] findCapitalCommitmentsByManangingCodeCapitalProjectId
  - [x] documented 
  - [x] implemented
- [x] findCommunityDistrictTiles
  - [x] documented 
  - [x] implemented
- [x] findCityCouncilDistrictTiles
  - [x] documented 
  - [x] implemented

TangoYankee:
@damonmcc I replaced the individual tickets with this meta ticket on the capital planning project. I thought that would help clean up that project board

pratishta:
[...]

PR #297, opened 2024-05-28
Revert "Add capital projects in community district endpoint"

TangoYankee:
Reverts NYCPlanning/ae-zoning-api#296

Until its order comes up in the Meta-ticket #278


TangoYankee:
Converting to draft while we decide whether to change to emoji driven documentation, which would obviate the need for this revert

PR #295, opened 2024-05-24
341 Document  Endpoint for City Council Districts

horatiorosa:
#### Description
Document City Council Districts endpoint

#### Tickets
 - Closes #241

PR #302, opened 2024-05-30
250 Implement findCityCouncilDistricts endpoint

### LINKED ISSUE #250: Implement findCityCouncilDistricts endpoint
## Acceptance criteria
- [x] Implement the `findCityCouncilDistricts` endpoint in compliance with #241
   - should be part of a `city-council-district` domain folder
- [x] Write end to end tests
- [x] Write unit tests

## helpful code: 
[the land use domain](https://github.com/NYCPlanning/ae-zoning-api/tree/main/src/land-use) 

## Related tickets
- blocked by #241

horatiorosa:
#### Description
implementation of city council district (ccd) endpoint
 - ccd controller, module, repository, service
 - cd unit and e2e tests
 - indentation fixes on openapi yaml
 - tweaks to README
 - regen ts and zod schemas

#### Tickets
 - Closes #250


README.md
```diff
@@ -90,6 +90,9 @@ Finally, to run the api locally:
 npm run dev
 ```
 
+- View the documentation locally through port 3000
```
pratishta, line 93: Maybe we can say localhost:3000 to keep it consistent with the previous localhost port

src/city-council-district/city-council-district.repository.ts
```diff
@@ -0,0 +1,24 @@
...
+    @Inject(DB)
+    private readonly db: DbType,
+  ) {}
+
+  async findMany(): Promise<FindManyRepo> {
+    try {
+      return await this.db.query.cityCouncilDistrict.findMany({
+        columns: {
+          id: true,
+          liFt: false,
```
pratishta, line 17: Outta curiosity, why is it that we need to specify false for `liFt` but not for `mercatorFill` or `mercatorLabel` just to get the `id`?

openapi/openapi.yaml
```diff
@@ -111,6 +111,30 @@ paths:
           $ref: "#/components/responses/NotFound"
         '500':
           $ref: "#/components/responses/InternalServerError"
+  /city-council-districts:
```
pratishta, line 114: This should not be part of the implementation PR. But once you pull changes from https://github.com/NYCPlanning/ae-zoning-api/pull/295, you should remove the 🚧 as part of this PR

pratishta (changes requested):

src/city-council-district/city-council-district.repository.ts
horatiorosa: It seems that without the `liFt` column, neither the `mercatorFill` or `mercatorLabel` are returned in the response. I am not sure why TBH. We can ask @TangoYankee for an explanation next week.

openapi/openapi.yaml
horatiorosa: yeah, this was only bc I was blocked without having the documentation so I branched off from #295 to create this one.

src/city-council-district/city-council-district.repository.ts
TangoYankee: The 'lift: false' is superfluous. specifying only 'id: true' will achieve the same result

horatiorosa: oh, i'll get rid of it (`liFt`)

horatiorosa: @pratishta **TIL**: it seems that when you specify what columns you want, the drizzle [docs](https://orm.drizzle.team/docs/rqb#partial-fields-select) state that ONLY those columns are returned. Not only that but when both true and false select options are present, all false options are ignored. So the `liFt: false` is bering ignored.

README.md
```diff
@@ -90,6 +90,9 @@ Finally, to run the api locally:
 npm run dev
 ```
 
+- View the documentation locally via `localhost:3000`
```
TangoYankee, line 93: Could we reorder this a little? It's clearer to me to first make a point that the application is running on port `3000`. After that first point, we can make these sub points. sub-point 1, the root route `/` has the documentation built from the point where the application was run. sub-point 2, the `/api/{endpoint}` routes have the actual implementations

compose.yaml
```diff
@@ -10,5 +10,9 @@ services:
       - POSTGRES_DB=${DATABASE_NAME}
     ports:
       - "8010:5432"
+    networks:
```
TangoYankee, line 13: These changes to compose should be their own PR- this docker network technique has technically been approved for the ETL.

TangoYankee (changes requested):
The implementation looks good. Next step would be to remove the construction emoji from the documentation because this PR implements it.

TangoYankee (approved):

pratishta (approved):

PR #303, opened 2024-06-04
246 Implement findCommunityDistrictsByBoroughId endpoint

### LINKED ISSUE #246: Implement findCommunityDistrictsByBoroughId endpoint
## Acceptance criteria
- [x] Implement the `findCommunityDistrictsByBoroughId` endpoint in compliance with #236
- [x] Write end to end tests
- [x] Write unit tests

## helpful code: 
[the zoning districts domain](https://github.com/NYCPlanning/ae-zoning-api/tree/main/src/zoning-district) 

## Related tickets
- blocked by #236

horatiorosa:
#### Description
implement find community districts by borough id endpoint
     -  cd by id controller, module, repository, service
     - added notNull constraints to community districts schema + migration
     - added unit and e2e tests
     
#### Tickets
Closes #246

### DIFF
--- a/src/borough/borough.service.spec.ts
+++ b/src/borough/borough.service.spec.ts
@@ -1,15 +1,19 @@
+import { BoroughRepository } from "src/borough/borough.repository";
+import { BoroughService } from "./borough.service";
 import { BoroughRepositoryMock } from "../../test/borough/borough.repository.mock";
 import { Test } from "@nestjs/testing";
-import { BoroughRepository } from "src/borough/borough.repository";
-import { findBoroughsQueryResponseSchema } from "src/gen";
-import { BoroughService } from "./borough.service";
+import {
+  findBoroughsQueryResponseSchema,
+  findCommunityDistrictsByBoroughIdQueryResponseSchema,
+} from "src/gen";
+import { ResourceNotFoundException } from "src/exception";
 
 describe("Borough service unit", () => {
   let boroughService: BoroughService;
 
-  beforeEach(async () => {
-    const boroughRepositoryMock = new BoroughRepositoryMock();
+  const boroughRepositoryMock = new BoroughRepositoryMock();
 
+  beforeEach(async () => {
     const moduleRef = await Test.createTestingModule({
       providers: [BoroughService, BoroughRepository],
     })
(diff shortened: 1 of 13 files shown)


src/borough/borough.service.ts
```diff
@@ -14,4 +15,18 @@ export class BoroughService {
...
   }
+
+  async findCommunityDistrictsByBoroughId(id: string) {
+    const boroughCheck = await this.boroughRepository.checkBoroughById(id);
+    if (boroughCheck === undefined) throw new ResourceNotFoundException();
+
+    const communityDistricts =
+      await this.boroughRepository.findCommunityDistrictsByBoroughId(id);
+
+    if (communityDistricts === undefined) throw new ResourceNotFoundException();
```
pratishta, line 26: if `findCommunityDistrictsByBoroughId` is defined in the repository as only being able to return `Promise<FindCommunityDistrictsByBoroughIdRepo>` and not `Promise<FindCommunityDistrictsByBoroughIdRepo | undefined>` can we still check for undefined here?

pratishta (changes requested):
We should remove 🚧 from `/boroughs/{boroughId}/community-districts:` in the openapi documentation since this PR implements it

src/borough/borough.service.ts
TangoYankee: > can we still check for undefined here

If you mean "we don't need to check for undefined", then I agree on removing this undefined check

horatiorosa: Oops, sorry! This was left over after I refactored the function DH initially wrote. It definitely should have been removed.

pratishta (approved):
If you could include 'Related to' or 'Closes' with the ticket number so the merge automatically handles the ticket, that'd be great. :) Otherwise LGTM

pratishta (changes requested):
Wait, wrong documentation... you want to remove the emoji from "summary: 🚧 Find community districts within a borough"

horatiorosa:
🤦🏽‍♂️ @pratishta

pratishta (approved):

src/borough/borough.service.spec.ts
```diff
@@ -20,8 +24,33 @@ describe("Borough service unit", () => {
...
+        await boroughService.findCommunityDistrictsByBoroughId(id);
+
+      expect(() =>
+        findCommunityDistrictsByBoroughIdQueryResponseSchema.parse(
+          communityDistricts,
+        ),
+      ).not.toThrow();
+    });
+
+    it("service should throw a resource error when requesting with a missing id", async () => {
```
TangoYankee, line 49: An empty string is different from an id that is not in the database. Empty strings are invalid request parameters. In practical terms, it would never make it to the service. The controller should throw an error before it gets to the service.

To test for a missing id, we want to look at what zod-mock lists and then pick an id that's not in that generated list. This is why we leverage seeding in our mocks- to ensure the generated data is always the same and never accidentally includes the id we designated as "missing"

horatiorosa: I misunderstood, thanks for explaining. I was thinking missing id as in not there and not as in missing in the db.

Issue #247, opened 2024-05-10
  Implement findCapitalProjectsByBoroughIdCommunityDistrictId endpoint 


TangoYankee:
## Acceptance criteria
- [x] Implement the `findCapitalProjectsByBoroughIdCommunityDistrictId` endpoint in compliance with #237
   - should be part of the `borough` domain folder
- [x] Write end to end tests
- [x] Write unit tests
- [x] Add geospatial index

## helpful code: 
[the tax lot domain](https://github.com/NYCPlanning/ae-zoning-api/tree/main/src/tax-lot) 

## Related tickets
- blocked by #237, #334

TylerMatteo:
Unfortunately didn't get too far with this but I did start a branch that someone can pick up: `247/projects-by-comm-district`. It has a commit for adding the constraints @TangoYankee mentions above.

TangoYankee:
@pratishta this ticket is affected by an incorrect geometry type. Community districts are supposed to be multipolygons. But, the 'lift' column is incorrectly typed as a 'point'. There are two fixes associated:
1) Fix the column itself
2) Remove the ["patch"](https://github.com/NYCPlanning/ae-data-flow/blob/capital-planning-flow/sql/populate_tables.sql#L478) in the data flow
    - This patch was put in to coerce the geometry into a point and just get the data flow working with the database until we could correct it.

pratishta:
Updated the draft PR with most of the work so far here: https://github.com/NYCPlanning/ae-zoning-api/pull/327#issuecomment-2177213282

PR #327, opened 2024-06-18
Implement find capital projects by community district endpoint

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/247

Several things to note:
- Wasn't sure where to place the check for community district ID. First commit (a2a9cd8247fa2dd5b09984b56c8c7f72f8089ba6) has it in `borough` and second commit (bda6221615735c7fca1b31d361a4ea7713652697) has it in its own domain `community-district`
- This endpoint is the first one where we're passing in both path and query params. I pass the `ZodTransformPipe` directly in the pram decorators because adding both param schemas in `UsePipes()` was causing errors.

### DIFF
--- a/src/borough/borough.controller.ts
+++ b/src/borough/borough.controller.ts
@@ -3,12 +3,17 @@ import {
   Get,
   Injectable,
   Param,
+  Query,
   UseFilters,
   UsePipes,
 } from "@nestjs/common";
 import { BoroughService } from "./borough.service";
 import {
+  FindCapitalProjectsByBoroughIdCommunityDistrictIdPathParams,
+  FindCapitalProjectsByBoroughIdCommunityDistrictIdQueryParams,
   FindCommunityDistrictsByBoroughIdPathParams,
+  findCapitalProjectsByBoroughIdCommunityDistrictIdPathParamsSchema,
+  findCapitalProjectsByBoroughIdCommunityDistrictIdQueryParamsSchema,
   findCommunityDistrictsByBoroughIdPathParamsSchema,
 } from "src/gen";
 import {
@@ -42,6 +47,26 @@ export class BoroughController {
   ) {
     return this.boroughService.findCommunityDistrictsByBoroughId(
       params.boroughId,
+    );
+  }
+
+  @Get("/:boroughId/community-districts/:communityDistrictId/capital-projects")
+  async findCapitalProjectsByBoroughIdCommunityDistrictId(
+    @Param(
+      new ZodTransformPipe(
+        findCapitalProjectsByBoroughIdCommunityDistrictIdPathParamsSchema,
+      ),
+    )
+    pathParams: FindCapitalProjectsByBoroughIdCommunityDistrictIdPathParams,
+    @Query(
+      new ZodTransformPipe(
+        findCapitalProjectsByBoroughIdCommunityDistrictIdQueryParamsSchema,
+      ),
+    )
+    queryParams: FindCapitalProjectsByBoroughIdCommunityDistrictIdQueryParams,
+  ) {
+    return this.boroughService.findCapitalProjectsByBoroughIdCommunityDistrictId(
+      { ...pathParams, ...queryParams },
     );
   }
 }
--- a/src/borough/borough.service.spec.ts
+++ b/src/borough/borough.service.spec.ts
@@ -1,24 +1,36 @@
 import { BoroughRepository } from "src/borough/borough.repository";
+import { CommunityDistrictRepository } from "src/community-district/community-district.repository";
 import { BoroughService } from "./borough.service";
 import { BoroughRepositoryMock } from "../../test/borough/borough.repository.mock";
 import { Test } from "@nestjs/testing";
 import {
   findBoroughsQueryResponseSchema,
+  findCapitalProjectsByBoroughIdCommunityDistrictIdQueryResponseSchema,
   findCommunityDistrictsByBoroughIdQueryResponseSchema,
 } from "src/gen";
 import { ResourceNotFoundException } from "src/exception";
+import { CommunityDistrictRepositoryMock } from "test/community-district/community-district.repository.mock";
 
 describe("Borough service unit", () => {
   let boroughService: BoroughService;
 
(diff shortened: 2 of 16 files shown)


pratishta:
I rebased off main with the Kubb upgrade changes. There are some things I did not have time for:
- More robust testing checking for query params e.g. default vs specified 
- Checking for community district ID... I wasn't sure if we wanted to have that function be dependent on checking the borough ID or if they should both be independent. I chose independent in this PR and this affects the repository mock functionality as well.
- Adding the ZodTransform pipe to `boroughs.controller.ts`... Adding the the new pipe causes some type errors in the param I couldn't quite figure out

test/borough/borough.repository.mock.ts
```diff
@@ -2,12 +2,17 @@ import {
   findManyRepoSchema,
   checkByIdRepoSchema,
   findCommunityDistrictsByBoroughIdRepoSchema,
+  findCapitalProjectsByBoroughIdCommunityDistrictIdRepoSchema,
 } from "src/borough/borough.repository.schema";
+// import { checkByCommunityDistrictIdRepoSchema, checkCommunityDistrictByIdMocks } from "src/community-district/community-district.repository.schema";
```
pratishta, line 7: Will remove this, clean up and squash once there's feedback about which wya to go for checking for community district ID

src/borough/borough.controller.ts
```diff
@@ -44,4 +49,24 @@ export class BoroughController {
       params.boroughId,
     );
   }
+
+  @Get("/:boroughId/community-districts/:communityDistrictId/capital-projects")
+  async findCapitalProjectsByBoroughIdCommunityDistrictId(
+    @Param(
```
TangoYankee, line 55: Oh, interesting. The way we solved this on Horatio's branch was to merge the definitions. Merging the defintions fits the prefixed pipes pattern better. However, it kinda waters down the schema definitions because it relies on "partial".

Using the pipe in each type of parameter might actually be more powerful because it uses the zod schema definition as-is

https://github.com/NYCPlanning/ae-zoning-api/blob/3a2ef3808bfd802fe4751b97aedbbbbdcfd39791/src/city-council-district/city-council-district.controller.ts#L57

test/borough/borough.repository.mock.ts
TangoYankee: Yeah, it seems to make more sense to have the "community district id check" be in the community district repository

TangoYankee (changes requested):
We'll also need correct the community district type to actually be polygons. Otherwise, we won't be getting the correct spatial intersections. Let me know if you need help with this or the corresponding data flow trickery.

https://github.com/NYCPlanning/ae-zoning-api/issues/247#issuecomment-2168735710

test/borough/borough.repository.mock.ts
```diff
@@ -2,12 +2,17 @@ import {
...
 import { generateMock } from "@anatine/zod-mock";
+import { CommunityDistrictRepositoryMock } from "test/community-district/community-district.repository.mock";
 
 export class BoroughRepositoryMock {
   numberOfMocks = 1;
 
+  communityDistrictRepoMock = new CommunityDistrictRepositoryMock();
```
TangoYankee, line 14: I would prefer we leveraged dependency injection instead of instantiating a new CommunityDistrictRepositoryMock object inside this mock.

It will also help with test consistency. The object instantiated here is different than the object instantiated directly in the test suite. It's only because we're using seeding that we get the same values. If we ever move away from seeding, we'll actually be getting different values

I made an example on a review branch: [247/ty/review](https://github.com/NYCPlanning/ae-zoning-api/commit/62c84a9ebef95cb93f6d4cf9882363c91c9fbea3)

src/borough/borough.service.spec.ts
```diff
@@ -53,4 +63,23 @@ describe("Borough service unit", () => {
...
+          communityDistrictId,
+        });
+
+      expect(() =>
+        findCapitalProjectsByBoroughIdCommunityDistrictIdQueryResponseSchema.parse(
+          capitalProjects,
+        ),
+      ).not.toThrow();
+    });
+  });
```
TangoYankee, line 84: There is room for better coverage of the behavior of the limit/offset/total/order.

See the city council district service test for an example suite. [findCapitalProjectsByCityCouncilId](https://github.com/NYCPlanning/ae-zoning-api/blob/46e1dbb52ec5b8443dcb3b804578c93391e5f567/src/city-council-district/city-council-district.service.spec.ts#L53)

TangoYankee (changes requested):
blocked by #334

test/borough/borough.repository.mock.ts
pratishta: Updated to inject the whole object to make a bit more explicit where `checkCommunityDistrictByIdMocks` is coming from.

src/borough/borough.controller.ts
pratishta: We might want to update CCD to directly use pipe in the param decorators so that both endpoints are consistent.

src/schema/community-district.ts
```diff
@@ -17,6 +17,7 @@ export const communityDistrict = pgTable(
   (table) => {
     return {
       pk: primaryKey({ columns: [table.boroughId, table.id] }),
+      liFtGix: index().using("GIST", table.liFt),
```
TangoYankee, line 20: We'll need a migration to actually apply this spatial index

src/borough/borough.controller.ts
TangoYankee: https://github.com/NYCPlanning/ae-zoning-api/issues/340

TangoYankee (changes requested):

src/schema/community-district.ts
TangoYankee: (Rebasing off main first, to make sure this branch also has all the existing migrations before creating new ones)

pratishta: Oop, thanks for catching. Updated

TangoYankee (approved):

Issue #334, opened 2024-06-27
Correct geometry type of community district long island state plane field


TangoYankee:
# Description
The `lift` field of the `community district` table is incorrectly typed as a multipoint. The correct geometry type is multipolygon. The schema needs to be corrected and a migration needs to be generated to apply the schema change. Because data already exists in the database, the migration will likely need to also drop the existing point data before changing the geometry type.

## Acceptance criteria 
- [x] The `liFt` geometry of the `community district` table is changed from a multipoint to a multipolygon
- [x] There is a migration to apply this schema change
- [x] The migration drops any existing data in the liFt column because they are multipoints, which is incorrect

PR #335, opened 2024-06-27
Update geom in community district schema and create migration file

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/334

### DIFF
--- a/db/migration/0013_petite_bulldozer.sql
+++ b/db/migration/0013_petite_bulldozer.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "community_district" DROP COLUMN IF EXISTS "li_ft";
--- a/db/migration/0014_strange_tarantula.sql
+++ b/db/migration/0014_strange_tarantula.sql
@@ -1,0 +1,1 @@
+ALTER TABLE "community_district" ADD COLUMN "li_ft" geometry(multiPolygon,2263);
--- a/db/migration/meta/0013_snapshot.json
+++ b/db/migration/meta/0013_snapshot.json
@@ -1,0 +1,1075 @@
+{
+  "id": "5248307d-5abe-4833-aea8-a9209bcef136",
+  "prevId": "b9e3c980-63df-44a1-9b88-aef74222e251",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.agency_budget": {
+      "name": "agency_budget",
+      "schema": "",
+      "columns": {
+        "code": {
+          "name": "code",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "type": {
+          "name": "type",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        },
+        "sponsor": {
+          "name": "sponsor",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": false
+        }
+      },
+      "indexes": {},
+      "foreignKeys": {
+        "agency_budget_sponsor_agency_initials_fk": {
+          "name": "agency_budget_sponsor_agency_initials_fk",
+          "tableFrom": "agency_budget",
+          "tableTo": "agency",
+          "columnsFrom": [
+            "sponsor"
+          ],
+          "columnsTo": [
+            "initials"
+          ],
+          "onDelete": "no action",
+          "onUpdate": "no action"
+        }
+      },
+      "compositePrimaryKeys": {},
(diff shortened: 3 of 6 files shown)


db/migration/0013_previous_lenny_balinger.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE "community_district" ALTER COLUMN "li_ft" SET DATA TYPE geometry(multiPolygon,2263);
```
pratishta, line 1: For "The migration drops any existing data in the liFt column because they are multipoints" is this something we'd manually add the SQL for here or is there a way to do it with Drizzle?

TangoYankee: I'll run this locally tomorrow. If it starts shouting errors, we'll have some answers

TangoYankee: Yes, it shouts errors- specifically about multipoints not being multipolygons. I'll share a code snippet for that.

TangoYankee, line 1:
```suggestion
UPDATE "community_district" SET "li_ft" = NULL; --> statement-breakpoint
ALTER TABLE "community_district" ALTER COLUMN "li_ft" SET DATA TYPE geometry(multiPolygon,2263);
```

pratishta: Updated. To confirm, when we want to do things like drop columns, we should manually edit the migration files?

TangoYankee: I find it's a grey area on whether it's better to create [a custom migration file](custom) from scratch or edit an auto-generated one. My guideline is that it's okay to edit the auto-generated migration as long as it doesn't contradict the snapshot meta-data.

It's important to highlight that we never [`DROP`](https://www.w3schools.com/SQL/sql_ref_drop_column.asp) the "li_ft" column. That would push us into "temporarily contradicting" the snapshot meta-data. We only update the content of the rows to remove any errant data that is incompatible with the incoming data structure.

TangoYankee (commented):
Holding off on approving until after demo at 2 (lest we realize a mistake too late)

TangoYankee (approved):

db/migration/0013_previous_lenny_balinger.sql
TylerMatteo: I strongly recommend against manually editing generated migrations. That's generally a big no-no for ORMs with migration generators and I don't see why Drizzle would be any different.

horatiorosa: Yeah, we should  only fix a migration with another migration.

TangoYankee: Can you expand on why it's a "no-no"?

pratishta: My initial hesitation was simply on the principle of not wanting to mess with automatically generated files. I'm unfamiliar with cross referencing the snapshot metadata. I see that we can use something like the `--custom` flag with Drizzle to generate an empty migration file for manual SQL (maybe same as what Tim linked, can't see it rn). I'm fine either way but would be good to agree on as a team for how we wanna deal

pratishta:
Updated to have two migration files, one custom to set incorrect point data to null and another generated from updating the community district geom type schema

TylerMatteo (dismissed):
Lgtm but maybe hold off on merging until an approval from @TangoYankee just to make sure we're all on the same page.

db/migration/0013_previous_lenny_balinger.sql
TangoYankee: @pratishta, I'm not sure whether you saw in chat. But, we were talking about actually dropping columns as we're doing two migrations anyway.

TangoYankee (changes requested):
Changing to two migrations

db/migration/0013_previous_lenny_balinger.sql
pratishta: Yup

TangoYankee: Did you have questions or concerns? It appears we're still "updating+altering" instead of "dropping+adding"

pratishta:
Updated to have two auto-generated migrations. First from deleting the li_ft column from the schema and generating. Second from adding the li_ft column with the polygon geometry type and generating again.

TangoYankee (approved):

Issue #342, opened 2024-07-12
Implement findCapitalProjectGeoJsonByManagingCodeCapitalProjectId endpoint


TylerMatteo:
## Acceptance Criteria

- [x] Implement `findCapitalProjectGeoJsonByManagingCodeCapitalProjectId` as per the OAS spec defined in #341, add tests.
- [x] Top-level geojson feature `id` should be equal the managing code and project id concatenated together into one string, similar to the [managingCodeCapitalProjectId](https://github.com/NYCPlanning/ae-zoning-api/blob/main/src/capital-project/capital-project.repository.ts#L86) property created for the tiles.
- [x] As outlined in the OAS spec, geometry type should be either `MultiPolygon` or `MultiPoint`, depending on the geometry for the given capital project.

## Blocked by
* #341

TangoYankee:
@TylerMatteo and @horatiorosa 

GeoJson coordinates are in "WGS84". However, the capital-project table only stores "Long Island State Plane" and "Web mercator" coordinates. We have two options to obtain the coordinates in "WGS84".

The first option is to convert from "Long Island State Plane" on each request for geojson. This is the simplest implementation because it requires no updates to the database structure or the supporting ETL (data-flow). It also saves on storage because it avoids an additional two columns of data (One for MultiPolygons and another for MultiPoints). It does require more computation because the geometries must be converted between coordinate systems on each request. Though, this should be a fairly lightweight computation.

The second option is to add columns for MultiPoints and MultiPolygons. This requires updating both the database schema to hold the data and the ETL to populate the data. The benefit is that we do the transformations once in the ETL and do not need to do them again on the requests.

PR #353, opened 2024-07-26
350/ccd geojson

### LINKED ISSUE #348: Document findCityCouncilDistrictGeoJsonByCityCouncilDistrctId endpoint
# Description

We need the geojson of each city council district. It will act as a mask when filtering capital projects within the district

## Acceptance criteria
- Document `findCityCouncilDistrictGeoJsonByCityCouncilDistrictId` endpoint
- Should use the `cityCouncilDistrictId` as the geojson `id`.
- Should use the `CityCouncilDistrict` schema as the geojson `properties`
- Should use the `MultiPolygon` schema as the geojson `geometry`
- Should use an enum with `Feature` as the option as the geojson `type`

### LINKED ISSUE #350: Implement findCityCouncilDistrictGeoJsonByCityCouncilDistrctId endpoint
# Acceptance Criteria
- [ ] `findCityCouncilDistrictGeoJsonByCityCouncilDistrctId` endpoint is implemented in compliance with OAS created in #348

TangoYankee:
# First ticket 
document the endpoint to return a geojson object for a city council district

closes #348

(Also remove the construction emoji for the implement ccd tile endpoint)

# Second ticket
implement find city council district geojson
    
implement the endpoint to retrieve the city council district geojson of a district specified by its id
    
closes #350


src/city-council-district/city-council-district.service.ts
```diff
@@ -28,6 +32,32 @@ export class CityCouncilDistrictService {
...
+  }: FindCityCouncilDistrictGeoJsonByCityCouncilDistrictIdPathParams) {
+    const cityCouncilDistrictGeoJson =
+      await this.cityCouncilDistrictRepository.findGeoJsonById({
+        cityCouncilDistrictId,
+      });
+    if (cityCouncilDistrictGeoJson === undefined)
+      throw new ResourceNotFoundException();
+
+    const geometry = JSON.parse(cityCouncilDistrictGeoJson.geometry);
+    const properties = produce(
```
TylerMatteo, line 46: Two things:
* I'm not sure this warrants adding immer as a dependency. This is API code so not concerned about bundle size, but concerned in terms of adding more code we depend on being maintained.
* When I pulled this branch `draft` shows up as a TS error because it is implicitly typed `any`. Is that happening for you? We should avoid that.

Maybe building `properties` manually isn't the worst thing for entities like this that don't have many properties? I don't think you would need this immer code or the `delete` if line 56 becomes
```
properties: { id: cityCouncilDistrictGeoJson.id},
```

src/city-council-district/city-council-district.service.ts
```diff
@@ -28,6 +32,32 @@ export class CityCouncilDistrictService {
...
+    cityCouncilDistrictId,
+  }: FindCityCouncilDistrictGeoJsonByCityCouncilDistrictIdPathParams) {
+    const cityCouncilDistrictGeoJson =
+      await this.cityCouncilDistrictRepository.findGeoJsonById({
+        cityCouncilDistrictId,
+      });
+    if (cityCouncilDistrictGeoJson === undefined)
+      throw new ResourceNotFoundException();
+
+    const geometry = JSON.parse(cityCouncilDistrictGeoJson.geometry);
```
TylerMatteo, line 45: Any reason you aren't doing `as MultiPolygon` here like we do in the tax lot service? I know it puts a type on the result that we don't really know for sure, but it does seem better than an implicit `any`.

TylerMatteo (changes requested):

src/city-council-district/city-council-district.service.ts
TangoYankee: I'll add it

src/city-council-district/city-council-district.service.ts
TangoYankee: For point one, I understand that it looks silly when there is only one property. However, we will have several endpoints returning GeoJSON with various properties. The properties of each endpoint is also prone to evolve. For example, we are implementing a Capital Project GeoJson endpoint for #342. This endpoint has several properties already. It also reasonable to anticipate the exact properties changing over time.

Using immer to separate the properties from the geometry gives us a simple rule that implements a robust,  low maintenance, and general solution.

TangoYankee: For point two, VSCode language server across both my Linux and Mac machines recognize it as `Partial<CityCouncilDistrictEntity`. Also, `npm run typecheck` does not flag it as an error. 

Have you run `npm i` on the branch to install `immer`?

TylerMatteo (dismissed):

TangoYankee:
Pratishta pointed out in a sister ticker that in the tax lot geojson we return only 6 decimal places instead of the default nine. https://github.com/NYCPlanning/ae-zoning-api/pull/356#discussion_r1700672022

This ticket has been updated to follow that precedent and only return 6 decimal places

TylerMatteo (approved):

PR #358, opened 2024-08-02
Resolve double counting commitments

### LINKED ISSUE #357: Capital Projects endpoint double counts commitments in "total"
## Description

The endpoint to return a single capital project returns a "commitmentsTotal" field. This field returns the sum of commitments for a project. Values for the commitments are accounted for at least twice, once as an individual item and again in the total. The "commitmentsTotal" is not filtering out the double counts, resulting in inflated return values. 

## Acceptance Criteria
- [x] The "commitmentsTotal" field of the "single capital project" endpoint only counts each commitment value once in its sum

TangoYankee:
Filter commitment funds to only the "totals" when returning a single capital project

closes #357


horatiorosa (approved):
1 + 1 = 3

✅

PR #371, opened 2024-09-03
368 Sort City Council District Find Many by Id

### LINKED ISSUE #368: Sort city council districts
# Description
It would be convenient if the `Select` inputs in the front end are sorted by district number in ascending order. We want to perform this sorting in the back end so that it is available to all applications that use the zoning api.

Image

## Acceptance criteria
- [x] The OpenAPI documentation is updated to with an 'order' variable that specifies that the city council districts are sorted by their IDs, as if they are numbers
- [x] The SQL/drizzle query for city council districts returns them sorted by their ids, as if they are numbers. (city council district '2' should come before '17'). 
  - Order is ascending.
- [x] The JSON response includes an `order` key with a value of `id`

horatiorosa:
#### Description 
- updated OpenAPI properties for CCD to include order variable
 - updated findMany to sort the CCDs by id as integers
 - updated CCD service spec to test `cityCouncilDistricts.order` 's value is `id`

#### Tickets
Closes #368

### DIFF
--- a/openapi/openapi.yaml
+++ b/openapi/openapi.yaml
@@ -311,8 +311,13 @@ paths:
                     type: array
                     items:
                       $ref: '#/components/schemas/CityCouncilDistrict'
+                  order: 
+                    type: string
+                    description: City council districts ids are sorted as if numbers in ascending order
+                    example: 'id'
                 required:
                  - cityCouncilDistricts
+                 - order
         '400':
           $ref: '#/components/responses/BadRequest'
         '500':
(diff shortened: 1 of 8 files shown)


openapi/openapi.yaml
```diff
@@ -1067,8 +1067,12 @@ components:
         id:
           type: string
           description: One or two character code to represent city council districts.
-          pattern: '^([0-9]{1,2})$' 
-          example: 25
+          pattern: '^([0-9]{1,2})$'
+          example: '25'
+        order:
```
TangoYankee, line 1072: `order` should be a parameter in the response for `city-council-districts`, rather than a parameter in the cityCouncilDistrict schema. 

When moving it to the response, we'll also want to add it to the list of required fields, otherwise we'll need to to tedious checks for the field every time we want to use it.

TangoYankee (changes requested):

openapi/openapi.yaml
horatiorosa: *property not parameter

TangoYankee (approved):
🚀

PR #375, opened 2024-09-10
204 Refactor: Split Open API documentation into multiple files

horatiorosa:
#### Description: 
Used redocly `split` to create a multi-filer structure by extracting parts into standalone separate `yaml`  files
   - component folder containing: 
     - parameters folder: contains all parameters 
     - responses folder: contains all responses 
     - schema folder: contains all schemas
   - paths folder: contains all paths
     - params in the path are indicated by curly brackets`{  }` 
     - backslashes ` / ` are replaced by underscores ` _ ` 
       - eg: `boroughs_{boroughId}_community-districts.yaml`


package.json
```diff
@@ -9,6 +9,7 @@
     "build": "npm run redoc:build && nest build",
     "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\" \"db/**/*.ts\"",
     "dev": "npm run redoc:build && nest start --watch",
+    "dev1": "npm run redoc:split",
```
TangoYankee, line 12: Remove before flight

TangoYankee (changes requested):
`dev1` is extra. Otherwise, this looks even better than what I had in mind

package.json
horatiorosa: woops. I meant to do that.

TangoYankee (approved):

PR #396, opened 2024-12-11
390 Remove enums from Zoning District Class

### LINKED ISSUE #390: Refactor zoning district class category from enum to check constraint
## Acceptance criteria
- [x] Remove the `categoryEnum` from `zoning-district-class`
- [x] Change the `category` field from a `categoryEnum` to a text type
- [x] Add a check constraint to the `category` field which ensures the value is one of the options previously enforced by the enum
  -  Note: the `zoningDistrictClassCategoryEntitySchema` should remain untouched. We are still using the field like an enum in the javascript api.
- [x] Generate a drizzle migration

## Related
Depends on #386

horatiorosa:
#### Description
- Removed the `categoryEnum` from `zoning-district-class`
 - Changed the category field from a `categoryEnum` to a `text`
 - Added a check constraint to the category field 

#### Tickets
Closes #390

### DIFF
--- a/src/zoning-district-class/zoning-district-class.repository.ts
+++ b/src/zoning-district-class/zoning-district-class.repository.ts
@@ -1,8 +1,9 @@
 import { Inject } from "@nestjs/common";
 import { DB, DbType } from "src/global/providers/db.provider";
-import { eq } from "drizzle-orm";
+import { eq, sql } from "drizzle-orm";
 import { DataRetrievalException } from "src/exception";
 import { zoningDistrictClass } from "src/schema";
+import { ZoningDistrictClassCategory } from "src/gen";
 import {
   FindManyRepo,
   FindByIdRepo,
@@ -17,7 +18,15 @@ export class ZoningDistrictClassRepository {
 
   async findMany(): Promise<FindManyRepo> {
     try {
-      return await this.db.query.zoningDistrictClass.findMany();
+      return await this.db
+        .select({
+          id: zoningDistrictClass.id,
+          description: zoningDistrictClass.description,
+          category: sql<ZoningDistrictClassCategory>`${zoningDistrictClass.category}`,
+          color: zoningDistrictClass.color,
+          url: zoningDistrictClass.url,
+        })
+        .from(zoningDistrictClass);
     } catch {
       throw new DataRetrievalException();
     }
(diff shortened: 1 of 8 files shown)


src/zoning-district-class/zoning-district-class.repository.ts
```diff
@@ -17,17 +18,44 @@ export class ZoningDistrictClassRepository {
...
+          category: true,
+          color: true,
+          url: true,
+        },
+      });
+
+      return result.map((zoningDistrictClass: any) => ({
+        ...zoningDistrictClass,
+        category: zoningDistrictClass.category as ZoningDistrictClassCategory,
+      }));
```
TangoYankee, line 34: The  `map` operation on line `31` adds an O(n) operation to achieve the type assertion. I think it would be more efficient to convert the `findMany` operation on line `21` to a `select` statement and perform the type coercion directly in there. That way, we achieve the same result in a single operation.

```typescript
return this.db.select([COLUMNS including casting the category]).from(zoningDistrictClass)...
```

src/zoning-district-class/zoning-district-class.repository.ts
```diff
@@ -17,17 +18,44 @@ export class ZoningDistrictClassRepository {
...
     } catch {
       throw new DataRetrievalException();
     }
   }
 
   async findById(id: string): Promise<FindByIdRepo | undefined> {
     try {
-      return await this.db.query.zoningDistrictClass.findFirst({
+      const result = await this.db.query.zoningDistrictClass.findFirst({
```
TangoYankee, line 42: Same feedback as above to use select statement.

When converting to a select statement, we can use `limit(1)` to achieve the same effect as `findFirst`. It will return an array of results. We will return either the first item in the array or `undefined` if there are no matching results.

src/zoning-district/zoning-district.repository.ts
```diff
@@ -61,7 +62,7 @@ export class ZoningDistrictRepository {
       return await this.db
         .select({
           id: zoningDistrictClass.id,
-          category: zoningDistrictClass.category,
+          category: sql<ZoningDistrictClassCategory>`${zoningDistrictClass.category}`,
```
TangoYankee, line 65: This is an example of the select syntax referenced by the other two comments.

TangoYankee (changes requested):

src/zoning-district-class/zoning-district-class.repository.ts
TangoYankee: > It will return an array of results. We will return either the first item in the array or undefined if there are no matching results.

Because we're changing to an array, we'll get a type error if we don't make some modifications. Here is a link to a code snippet that should resolves these.

https://github.com/NYCPlanning/ae-zoning-api/commit/4d7666117a1d00b1363baec36ea6acdacbe76a57

TangoYankee (approved):

Issue #404, opened 2025-01-21
Add new capital projects endpoint


dhochbaum-dcp:
Endpoint should be `/capital-projects`.

- [x] Endpoint should have the following filters as query parameters:
   - [x] limit (1-100, default 20)
   - [x] offset (>=0, default 0)

- [x] Endpoint should be reflected in OpenAPI spec (regenerate files as needed)
- [x] Add tests for new endpoint

TangoYankee:
I think this ticket should be limited in scope to only the query parameters that already exist on the production capital-project endpoints, `limit` and `offset`

pratishta:
Can we get some rationale for why we want to return capital projects under the CD and CCD endpoints as well as its own endpoint? And add filters to all of them? I'm trying to get some understanding as to why our API should care about capital projects within a specified geography, and why adding a geospatial filter to `capital-projects` might not make sense. 

My current thought process is, yeah it's better to maintain fewer endpoints if they're redundant. But also maybe on the other hand, it's nice to have a robust API for different needs. Specifically I'm not sure what @TylerMatteo meant with 'hierarchical relationship'

TangoYankee:
I started a discussion on when we might want to use path and query parameters, as they apply to capital projects
https://github.com/NYCPlanning/ae-zoning-api/discussions/406

PR #410, opened 2025-01-30
404 Add new capital projects endpoint

dhochbaum-dcp:
Endpoint should be `/capital-projects`.

- [x] Endpoint should have the following filters as query parameters:
   - [x] limit (1-100, default 20)
   - [x] offset (>=0, default 0)

- [x] Endpoint should be reflected in OpenAPI spec (regenerate files as needed)
- [x] Add tests for new endpoint

Completes #404

### DIFF
--- a/openapi/paths/capital-projects.yaml
+++ b/openapi/paths/capital-projects.yaml
@@ -1,0 +1,21 @@
+get:
+  summary: Find paginated capital projects.
+  operationId: findCapitalProjects
+  tags:
+    - Capital Projects
+  parameters:
+    - $ref: ../components/parameters/limitParam.yaml
+    - $ref: ../components/parameters/offsetParam.yaml
+  responses:
+    '200':
+      description: >-
+        An object containing pagination metadata and an array of capital
+        projects
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/CapitalProjectPage.yaml
+    '400':
+      $ref: ../components/responses/BadRequest.yaml
+    '500':
+      $ref: ../components/responses/InternalServerError.yaml
--- a/src/capital-project/capital-project.controller.ts
+++ b/src/capital-project/capital-project.controller.ts
@@ -2,6 +2,7 @@ import {
   Controller,
   Get,
   Param,
+  Query,
   Res,
   UseFilters,
   UsePipes,
@@ -12,6 +13,7 @@ import {
   FindCapitalProjectByManagingCodeCapitalProjectIdPathParams,
   FindCapitalProjectGeoJsonByManagingCodeCapitalProjectIdPathParams,
   FindCapitalProjectTilesPathParams,
+  FindCapitalProjectsQueryParams,
   findCapitalCommitmentsByManagingCodeCapitalProjectIdPathParamsSchema,
   findCapitalProjectByManagingCodeCapitalProjectIdPathParamsSchema,
   findCapitalProjectGeoJsonByManagingCodeCapitalProjectIdPathParamsSchema,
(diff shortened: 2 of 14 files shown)


openapi/paths/capital-projects.yaml
```diff
@@ -0,0 +1,23 @@
...
+        An object containing pagination metadata and an array of capital
+        projects
+      content:
+        application/json:
+          schema:
+            $ref: ../components/schemas/CapitalProjectPage.yaml
+    '400':
+      $ref: ../components/responses/BadRequest.yaml
+    '404':
+      $ref: ../components/responses/NotFound.yaml
```
TangoYankee, line 21: The capital projects endpoint does not currently have a case where it 404s. (And I think the upcoming query parameter validations will also only cause 400)

```suggestion
    '404':
      $ref: ../components/responses/NotFound.yaml
```

src/capital-project/capital-project.service.ts
```diff
@@ -19,6 +20,16 @@ export class CapitalProjectService {
...
   ) {}
 
+  async findCapitalProjects({
+    limit = 20,
+    offset = 0,
+  }: FindCapitalProjectsQueryParams) {
+    return await this.capitalProjectRepository.findCapitalProjects({
+      limit,
+      offset,
+    });
```
TangoYankee, line 30: Before returning the result of the query, we need to place the results in an object with the metadata about order and pagination

test/capital-project/capital-project.repository.mock.ts
```diff
@@ -16,6 +17,27 @@ import {
...
+  );
+
+  async findCapitalProjects() {
+    return {
+      limit: 20,
+      offset: 0,
+      total: 4,
+      order: "managingCode, capitalProjectId",
+      capitalProjects: this.findCapitalProjectsMocks,
+    };
```
TangoYankee, line 38: The repository mock reflects what the database query would return, which does not include the order and pagination metadata. (These data are added by the service)

```suggestion
    return this.findCapitalProjectsMocks;
```

src/capital-project/capital-project.controller.ts
```diff
@@ -33,6 +36,15 @@ import { ZodTransformPipe } from "src/pipes/zod-transform-pipe";
 export class CapitalProjectController {
   constructor(private readonly capitalProjectService: CapitalProjectService) {}
 
+  @UsePipes(new ZodTransformPipe(findCapitalProjectsQueryParamsSchema))
+  @Get("/")
+  async findCapitalProjects(
```
TangoYankee, line 41: Okay, I know this discussion is buried and following it is tedious ...but...naming conventions on controllers/controllers/and repositories...

[Discussion on naming conventions](https://github.com/NYCPlanning/ae-zoning-api/discussions/115#discussion-5982504)

In this case, the guidelines would say to use `findMany` instead of `findCapitalProjects`. This applies to the controller, service, and repository.

@TylerMatteo , I'm not sure how much you think we should enforce this.

src/capital-project/capital-project.controller.ts
```diff
@@ -33,6 +36,15 @@ import { ZodTransformPipe } from "src/pipes/zod-transform-pipe";
 export class CapitalProjectController {
   constructor(private readonly capitalProjectService: CapitalProjectService) {}
 
+  @UsePipes(new ZodTransformPipe(findCapitalProjectsQueryParamsSchema))
```
TangoYankee, line 39: I think we're double parsing the query parameters. Once on this line and again on line 42. 

Either approach works but we should only keep one.

TangoYankee:
In addition to the service test, we should also have end to end tests

horatiorosa (approved):
Looks good y'all.

🚀

PR #405, opened 2025-01-22
400/add managing agency filter type to capital projects

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/400

PR #411, opened 2025-02-03
407&408 Add community & city council district filter to capital projects endpoint

### LINKED ISSUE #407: Add community district filter to capital projects endpoint
### Acceptance Criteria

- [ ] The `/capital-projects` endpoint has an optional `communityDistrictId` query parameter
  - ***Note**: a communityDistrictId path parameter already exists. It is okay for the query to share the name because it is placed in a different object. However, we do need a way to differentiate the file names. We may want to refactor the parameter files in a future ticket to end with "Path" when they're path parameters and "Query" when they're query parameters*
- [ ] The query parameter accepts the three character concatenation of the borough and community district ids
- [ ] The capital-projects endpoint must only return projects that spatially intersect with the specified community district
- [ ] New query parameter is reflected in OpenAPI spec (regenerate files as needed)
  - Description should include instructions to concatenate the borough and community district id
- [ ] Add tests for new query parameter

### Validation
- [ ] Return a 400 (invalid request parameter exception) if given ID does structure does not follow valid structure of community district id
- [ ] Return a 400 (invalid request parameter exception) if no community district with given ID is found

### LINKED ISSUE #408: Add city council district filter to capital projects endpoint
### Acceptance Criteria

- [ ] The `/capital-projects` endpoint has an optional `cityCouncilDistrictId` query parameter
  - ***Note**: a cityCouncilDistrictId path parameter already exists. It is okay for the query parameter to share the name because it is placed in a different object. However, we do need a way to differentiate the file names. We may want to refactor the parameter files in a future ticket to end with "Path" when they're path parameters and "Query" when they're query parameters*
- [ ] The capital-projects endpoint must only return projects that spatially intersect with the specified city council district
- [ ] New query parameter is reflected in OpenAPI spec (regenerate files as needed)
- [ ] Add tests for new query parameter

### Validation
- [ ] Return 400 (invalid request parameter) if no city council district district with the given ID exists
- [ ] Return 400 (invalid request parameter) if given ID does not follow the valid pattern for city council district ids

dhochbaum-dcp:
Closes #407 
Closes #408

### DIFF
--- a/openapi/components/parameters/ccdParam.yaml
+++ b/openapi/components/parameters/ccdParam.yaml
@@ -5,4 +5,5 @@ schema:
   type: string
   pattern: ^([0-9]{1,2})$
   example: '25'
+  nullable: true
 description: One or two character code to represent city council districts.
--- a/openapi/components/parameters/ccdParam.yaml
+++ b/openapi/components/parameters/ccdParam.yaml
@@ -1,0 +1,8 @@
+name: ccd
+required: false
+in: query
+schema:
+  type: string
+  pattern: ^([0-9]{1,2})$
+  example: '25'
+description: One or two character code to represent city council districts.
--- a/openapi/components/parameters/cityCouncilDistrictIdQueryParam.yaml
+++ b/openapi/components/parameters/cityCouncilDistrictIdQueryParam.yaml
@@ -5,5 +5,4 @@ schema:
   type: string
   pattern: ^([0-9]{1,2})$
   example: '25'
-  nullable: true
 description: One or two character code to represent city council districts.
--- a/openapi/components/parameters/cityCouncilDistrictIdQueryParam.yaml
+++ b/openapi/components/parameters/cityCouncilDistrictIdQueryParam.yaml
@@ -1,4 +1,4 @@
-name: ccd
+name: cityCouncilDistrictId
 required: false
 in: query
 schema:
--- a/openapi/components/parameters/communityDistrictIdQueryParam.yaml
+++ b/openapi/components/parameters/communityDistrictIdQueryParam.yaml
@@ -1,4 +1,4 @@
-name: cd
+name: communityDistrictId
 required: false
 in: query
 schema:
--- a/openapi/components/parameters/communityDistrictIdQueryParam.yaml
+++ b/openapi/components/parameters/communityDistrictIdQueryParam.yaml
@@ -5,6 +5,5 @@ schema:
   type: string
   pattern: ^([0-9]{3})$
   example: '101'
-  nullable: true
 description: >-
   The three character numeric string containing the concatenation of the borough and community district ids.
--- a/src/capital-project/capital-project.repository.ts
+++ b/src/capital-project/capital-project.repository.ts
@@ -34,15 +34,15 @@ export class CapitalProjectRepository {
   ) {}
 
   async findMany({
-    ccd,
-    cd,
-    boro,
+    cityCouncilDistrictId,
+    communityDistrictId,
+    boroughId,
     limit,
     offset,
   }: {
-    ccd: string | null;
-    cd: string | null;
-    boro: string | null;
+    cityCouncilDistrictId: string | null;
+    communityDistrictId: string | null;
+    boroughId: string | null;
     limit: number;
     offset: number;
   }): Promise<FindCapitalProjectsRepo> {
@@ -72,11 +72,13 @@ export class CapitalProjectRepository {
         )
         .where(
           and(
-            ccd !== null ? eq(cityCouncilDistrict.id, ccd) : undefined,
-            cd !== null && boro !== null
+            cityCouncilDistrictId !== null
+              ? eq(cityCouncilDistrict.id, cityCouncilDistrictId)
+              : undefined,
+            communityDistrictId !== null && boroughId !== null
               ? and(
-                  eq(communityDistrict.boroughId, boro),
-                  eq(communityDistrict.id, cd),
+                  eq(communityDistrict.boroughId, boroughId),
+                  eq(communityDistrict.id, communityDistrictId),
                 )
               : undefined,
           ),
--- a/src/capital-project/capital-project.repository.ts
+++ b/src/capital-project/capital-project.repository.ts
@@ -39,8 +39,8 @@ export class CapitalProjectRepository {
     limit,
     offset,
   }: {
-    ccd: string;
-    cd: string;
+    ccd: string | null;
+    cd: string | null;
     limit: number;
     offset: number;
   }): Promise<FindCapitalProjectsRepo> {
@@ -70,8 +70,8 @@ export class CapitalProjectRepository {
         )
         .where(
           and(
-            ccd ? eq(cityCouncilDistrict.id, ccd) : undefined,
-            cd
+            ccd !== null ? eq(cityCouncilDistrict.id, ccd) : undefined,
+            cd !== null
               ? and(
                   eq(communityDistrict.boroughId, parseInt(cd[0]).toString()),
                   eq(
--- a/src/capital-project/capital-project.repository.ts
+++ b/src/capital-project/capital-project.repository.ts
@@ -36,11 +36,13 @@ export class CapitalProjectRepository {
   async findMany({
     ccd,
     cd,
+    boro,
     limit,
     offset,
   }: {
     ccd: string | null;
     cd: string | null;
+    boro: string | null;
     limit: number;
     offset: number;
   }): Promise<FindCapitalProjectsRepo> {
@@ -71,13 +73,10 @@ export class CapitalProjectRepository {
         .where(
           and(
             ccd !== null ? eq(cityCouncilDistrict.id, ccd) : undefined,
-            cd !== null
+            cd !== null && boro !== null
               ? and(
-                  eq(communityDistrict.boroughId, parseInt(cd[0]).toString()),
-                  eq(
-                    communityDistrict.id,
-                    parseInt(cd.slice(1, 3)).toString().padStart(2, "0"),
-                  ),
+                  eq(communityDistrict.boroughId, boro),
+                  eq(communityDistrict.id, cd),
                 )
               : undefined,
           ),
--- a/src/capital-project/capital-project.repository.ts
+++ b/src/capital-project/capital-project.repository.ts
@@ -15,6 +15,8 @@ import {
   capitalCommitment,
   capitalCommitmentFund,
   capitalProject,
+  cityCouncilDistrict,
+  communityDistrict,
 } from "src/schema";
 import {
   CheckByManagingCodeCapitalProjectIdRepo,
@@ -32,9 +34,13 @@ export class CapitalProjectRepository {
   ) {}
 
   async findMany({
+    ccd,
+    cd,
     limit,
     offset,
   }: {
+    ccd: string;
+    cd: string;
     limit: number;
     offset: number;
   }): Promise<FindCapitalProjectsRepo> {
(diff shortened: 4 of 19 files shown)


openapi/components/parameters/ccdParam.yaml
```diff
@@ -0,0 +1,8 @@
+name: ccd
```
TangoYankee, line 1: I think we should avoid abbreviating these parameters. As time passes, we'll add more domains and spend more time context switching; it will be hard to remember `ccd` and `cd` by looking at them. Writing them out as `cityCouncilDistrict` and `communityDistrict` will make it easier to understand them at a glance.

src/capital-project/capital-project.service.ts
```diff
@@ -20,8 +20,15 @@ export class CapitalProjectService {
     private readonly capitalProjectRepository: CapitalProjectRepository,
   ) {}
 
-  async findMany({ limit = 20, offset = 0 }: FindCapitalProjectsQueryParams) {
+  async findMany({
+    limit = 20,
+    offset = 0,
+    ccd = "",
+    cd = "",
```
TangoYankee, line 27: Similar feedback to the [managing agency PR](https://github.com/NYCPlanning/ae-zoning-api/pull/405#discussion_r1928897633). In sort, using `null` as value to represent the lack of `district`. Then in the repository, specifically checking for `null` rather than a falsy (empty string) value.  

```suggestion
    ccd = null,
    cd = null,
```

openapi/components/parameters/ccdParam.yaml
dhochbaum-dcp: I disagree, for three reasons:
1) We consistently use the abbreviations `nta`, `cdta`, `cd`, and `ccd` throughout our projects, so it does not cause issues with context switching.
2) There is value in having shorter urls.
3) I believe we should be consistent in using the full names of variables when using full values, and acronyms for variable names when we are using shortened values.  For example:
 - `borough` should be used for "Manhattan", while `boro` should be used when the value is "MN" or "3"
 - `communityDistrict` should be used when the value is "Manhattan 11", and `cd` should be used when the value is "MN11" or "301"

src/capital-project/capital-project.service.ts
dhochbaum-dcp: Done.

src/capital-project/capital-project.repository.ts
```diff
@@ -50,6 +56,32 @@ export class CapitalProjectRepository {
...
+        )
+        .where(
+          and(
+            ccd ? eq(cityCouncilDistrict.id, ccd) : undefined,
+            cd
+              ? and(
+                  eq(communityDistrict.boroughId, parseInt(cd[0]).toString()),
+                  eq(
+                    communityDistrict.id,
+                    parseInt(cd.slice(1, 3)).toString().padStart(2, "0"),
```
TangoYankee, line 79: Separating the borough and community portions of the community district id should be done in the service, rather than in the sql query. I also think we can simplify the parsing. 

By switching our "no community district" representation to a null value, we know that if the community district is present then it must be a string with three characters. (The Zod parsing guarantees this). With this guarantee, we can simplify the extracting to:

```typescript
const boroughId = cd.slice(0,1)
const communityDistrictId = cd.(1, 3)
```

By moving this extraction to the service, we separate the concerns so that the repository/query doesn't need to know about the parsing. It also means that if there is an error while parsing, it is not obscured by our catch all "data retrieval error"

src/capital-project/capital-project.repository.ts
```diff
@@ -50,6 +56,32 @@ export class CapitalProjectRepository {
...
+        )
+        .leftJoin(
+          communityDistrict,
+          sql`
+          ST_Intersects(${communityDistrict.liFt}, ${capitalProject.liFtMPoly})
+          OR ST_Intersects(${communityDistrict.liFt}, ${capitalProject.liFtMPnt})`,
+        )
+        .where(
+          and(
+            ccd ? eq(cityCouncilDistrict.id, ccd) : undefined,
```
TangoYankee: More explicit check for missing district values: (repeat for community district + borough Id)

```suggestion
            ccd !== null ? eq(cityCouncilDistrict.id, ccd) : undefined,
```

src/capital-project/capital-project.service.ts
```diff
@@ -20,8 +20,15 @@ export class CapitalProjectService {
...
+  }: FindCapitalProjectsQueryParams) {
     const capitalProjects = await this.capitalProjectRepository.findMany({
+      ccd,
+      cd,
       limit,
       offset,
     });
```
TangoYankee, line 34: The acceptance criteria also specify that we will check whether the requested districts actually exist, or else return a 400. We can check both districts in parallel by placing their requests in a list of promises and waiting for them with `Promise.all`. We should be able to reuse the [`checkCityCouncilDistrictById`](https://github.com/NYCPlanning/ae-zoning-api/blob/9bd287e21c69da66c848d7223605d1241eeb5fdd/src/city-council-district/city-council-district.repository.ts#L36) and [`checkCommunityDistrictById`](https://github.com/NYCPlanning/ae-zoning-api/blob/9bd287e21c69da66c848d7223605d1241eeb5fdd/src/community-district/community-district.repository.ts#L28) functions in their respective repositories. Though now that I'm looking at `checkCommunityDistrictById`, I believe there is a logical error; the community districts are only unique within their borough. The check should look at the borough and community district Ids in the same query.

TangoYankee (changes requested):
I am concerned about the lack of process around coming to a agreement

dhochbaum-dcp:
> I am concerned about the lack of process around coming to a agreement

On that, we agree.

openapi/components/parameters/ccdParam.yaml
TylerMatteo: Let's prefer writing out params over shortening or abbreviating, with exceptions for certain tokens like `Id`. I think it makes the API more readable at a glance and I've been on projects in the past where you burn cycles as a new developer remembering how past devs decided to abbreviate things. We'll likely run into situations in the future where writing out becomes impractical but preferring writing out is a good default best practice and I think it makes sense for the situation here.

As for the `borough` use case, I would probably expect something like `boroughId` or `boroughCode`? That way we're being explicit about what property of the borough we're using for the look up, just like with `cityCouncilDistrictId`, etc.

src/capital-project/capital-project.service.ts
dhochbaum-dcp: Checks to make sure the districts exist have been added, and `checkCommunityDistrictById` has been modified to check boroughId and communityDistrictId.

openapi/components/parameters/communityDistrictIdQueryParam.yaml
```diff
@@ -0,0 +1,10 @@
+name: communityDistrictId
+required: false
+in: query
+schema:
+  type: string
+  pattern: ^([0-9]{3})$
+  example: '101'
+  nullable: true
```
TylerMatteo, line 8: This shouldn't be `nullable`. Removing this line will change it to the default value of `false`. Having `null` as a default value in the service is fine, but the actual query param isn't `nullable`.

openapi/components/parameters/cityCouncilDistrictIdQueryParam.yaml
```diff
@@ -0,0 +1,9 @@
+name: cityCouncilDistrictId
+required: false
+in: query
+schema:
+  type: string
+  pattern: ^([0-9]{1,2})$
+  example: '25'
+  nullable: true
```
TylerMatteo, line 8: Same as below

openapi/components/parameters/communityDistrictIdQueryParam.yaml
dhochbaum-dcp: This conflicts with Tim's request [above](https://github.com/NYCPlanning/ae-zoning-api/pull/411#discussion_r1941383063).

Changing this changes the `FindCapitalProjectsQueryParams` type to disallow the option for `communityDistrictId`  from being `null`, which then prevents us from using `null` to represent the lack of `communityDistrictId`.

openapi/components/parameters/cityCouncilDistrictIdQueryParam.yaml
dhochbaum-dcp: This conflicts with Tim's request [above](https://github.com/NYCPlanning/ae-zoning-api/pull/411#discussion_r1941383063).

Changing this changes the `FindCapitalProjectsQueryParams` type to disallow the option for `cityCouncilDistrictId` from being `null`, which then prevents us from using `null` to represent the lack of `cityCouncilDistrictId`.

src/capital-project/capital-project.service.ts
```diff
@@ -8,20 +8,58 @@ import {
...
+    cityCouncilDistrictId !== null &&
+      checklist.push(
+        this.cityCouncilDistrictRepository.checkCityCouncilDistrictById(
+          cityCouncilDistrictId,
+        ),
+      );
+    communityDistrictId !== null &&
+      checklist.push(
+        this.communityDistrictRepository.checkCommunityDistrictById(
+          communityDistrictId.slice(0, 1),
```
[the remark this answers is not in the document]
dhochbaum-dcp: Because `communityDistrictId` is possibly `null`, that would require declaring the `const`s like so:
```
const borough = communityDistrictId === null ? "CANNOT USE NULL" : communityDistrictId.slice(0, 1);
const communityDistrict = communityDistrictId === null ? "NEEDS TO BE A STRING" : communityDistrictId.slice(1, 3);
```
Before I go ahead with this, are you sure doing `null` checks and assigning strings never to be used is preferable to the current implementation?

src/capital-project/capital-project.service.spec.ts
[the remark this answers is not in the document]
dhochbaum-dcp: Added.

src/capital-project/capital-project.service.spec.ts
```diff
@@ -1,7 +1,12 @@
 import { CapitalProjectRepositoryMock } from "test/capital-project/capital-project.repository.mock";
```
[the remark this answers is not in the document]
dhochbaum-dcp: I think with everything this will require to do properly, it should be broken out into a separate issue.  

`findCapitalProjectsByBoroughIdCommunityDistrictIdMocks` in `borough.repository.mock.ts` is not set up correctly.  It returns an object where the key is a `communityDistrictId`, and the value is an array of mock capital projects.  This is then used by `findCapitalProjectsByBoroughIdCommunityDistrictId`, which only checks against the `communityDistrictId` and not the combination of `boroughId` and `communityDistrictId`, because the `boroughId` is never generated or used in `findCapitalProjectsByBoroughIdCommunityDistrictIdMocks`.  (This mimics the [issue Tim mentioned](https://github.com/NYCPlanning/ae-zoning-api/pull/411#discussion_r1941425626) with `checkCommunityDistrictById` which I have corrected.

`city-council-district.repository.mock.ts` does not contain a `findCapitalProjectsByCityCouncilDistrictIdMocks` function, and I assume you also want that to be added.

I'm also confused about how you would like to be able to scale the `{ communityDistrictId: [{capitalProject}, {capitalProject}, ...] }` logic for the `findMany()` to deal with multiple parameters by which to filter.  How do we mimic a query where someone selects a `communityDistrictId` and `cityCouncilDistrictId`, or adds in a `managingAgency`, or  `projectType`, `commitmentsTotalMin` and/or `commitmentsTotalMax`?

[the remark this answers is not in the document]
TangoYankee: I made a branch that leverages a tuple to mimic criteria being associated with certain capital projects. 

https://github.com/NYCPlanning/ae-zoning-api/pull/417

(It's a draft PR against main, rather than a PR against this branch because it also flattens the change history in a single commit based off main.)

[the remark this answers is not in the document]
dhochbaum-dcp: Great, thank you.  Please mark it ready for review.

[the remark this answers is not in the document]
TangoYankee: I have updated the PR as ready to review

PR #415, opened 2025-02-10
Fix community district check

### LINKED ISSUE #414: Bug: Check of community district id
## Description

The check for the existence a community district id only looks at the `id` of portion, without simultaneously checking the `boroughId` portion. The check needs to look at both portions simultaneously, otherwise the key is not guaranteed to be unique. There is a also a capital project service that looks at the borough and id independently, creating a subtle logical bug.

## Acceptance criteria
- [ ] The check for community districts looks at the id and borough id simultaneously
- [ ] The capital project service is updated to look at this community district check
- [ ] Tests are updated

TangoYankee:
check borough and community district ids together

closes #414

### DIFF
--- a/src/borough/borough.service.spec.ts
+++ b/src/borough/borough.service.spec.ts
@@ -104,10 +104,9 @@ describe("Borough service unit", () => {
   });
 
   describe("findCapitalProjectsByBoroughIdCommunityDistrictId", () => {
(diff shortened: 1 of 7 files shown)


dhochbaum-dcp (approved):
Nice work!

PR #428, opened 2025-03-05
Commitment totals in capital project tiles

### LINKED ISSUE #426: Add `commitmentsTotal` value to Capital Project tile endpoints
## Acceptance Criteria

- [ ] Add new `commitmentsTotal` property to data returned by all Capital Project tile endpoints
- [ ] Value should be numeric and equal the total dollar amount for all commitments associated with the given project. For any given project, it should equal the `commitmentsTotal` field [already included](https://github.com/NYCPlanning/ae-zoning-api/blob/main/src/capital-project/capital-project.repository.ts#L143) in the JSON Capital Projects endpoints.

TangoYankee:
# Description

Add total value of commitments to tiles for capital projects

closes #426

## Test plan
Run with this front end code to see the commitmentsTotal logged to the browser console: [test/426](https://github.com/NYCPlanning/ae-cp-map/tree/test/426)

### DIFF
--- a/src/borough/borough.repository.ts
+++ b/src/borough/borough.repository.ts
@@ -9,8 +9,13 @@ import {
   FindCommunityDistrictGeoJsonByBoroughIdCommunityDistrictIdRepo,
   FindCapitalProjectTilesByBoroughIdCommunityDistrictIdRepo,
 } from "./borough.repository.schema";
-import { capitalProject, communityDistrict } from "src/schema";
-import { eq, sql, and, isNotNull, asc } from "drizzle-orm";
+import {
+  capitalCommitment,
+  capitalCommitmentFund,
+  capitalProject,
+  communityDistrict,
+} from "src/schema";
+import { eq, sql, and, isNotNull, asc, sum } from "drizzle-orm";
 import {
   CapitalProjectCategory,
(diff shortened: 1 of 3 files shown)


src/borough/borough.repository.ts
```diff
@@ -183,12 +191,31 @@ export class BoroughRepository {
...
+          eq(capitalCommitmentFund.capitalCommitmentId, capitalCommitment.id),
+        )
         .where(
           and(
             eq(communityDistrict.id, communityDistrictId),
             eq(communityDistrict.boroughId, boroughId),
+            eq(capitalCommitmentFund.category, "total"),
           ),
         )
+        .groupBy(
```
TylerMatteo, line 212: What's the purpose of this `groupBy`? I wouldn't necessarily expect to see this new clause for adding a property to the tiles.

TangoYankee: Without the groupBy, we would get an entry for every commitment that belongs to the capital project. Also, the value of the "commitmentsTotal" property would be the total for only that commitment.

With the groupBy, we get a single entry for each capital project. Also, the value of the "commitmentsTotal" property is the summation of all commitments for that capital project.

TylerMatteo (approved):

PR #434, opened 2025-03-20
Issue 432 Conditional Joins for Conditional Filters

### LINKED ISSUE #432: Conditional joins for conditional filters
# Description
We have multiple filters for the capital projects endpoint. Not every filter is always supplied. Consequently, we conditionally apply the filter in the `where` clause. Some of these filters require joining to another table. These tables are always run, regardless of whether the corresponding filter is requested. This results in many unnecessary joins. We should also apply a condition to joins. 

example conditional join:
[ty/exp/count-in-cache](https://github.com/NYCPlanning/ae-zoning-api/compare/main...ty/exp/count-in-cache#diff-d03a7cb979100a5a6b6f454a541fd6bfc32258c0612c835a85477d8e154e8cb7R101-R106)
```typescript
leftJoin(
          communityDistrict,
          and(
            sql`${communityDistrictId !== null && boroughId !== null} IS TRUE`,
            or(
              sql`ST_Intersects(${communityDistrict.liFt}, ${capitalProject.liFtMPoly})`,
              sql`ST_Intersects(${communityDistrict.liFt}, ${capitalProject.liFtMPnt})`,
            ),
          ),
        )
```
 
# Acceptance criteria
- [ ] Conditional statements are applied to joins in `findMany` method of `capital-projects.repository`

horatiorosa:
#### Description
Refactored the `findMany` method in the capital projects endpoint
 -  joins had a 'circuit-breaker" based on whether or not the primary keys of the join are null
   - if the primary keys were null, skip the db query, otherwise query away.

#### TIckets
Closes #432

### DIFF
--- a/src/capital-project/capital-project.repository.ts
+++ b/src/capital-project/capital-project.repository.ts
@@ -1,5 +1,5 @@
 import { Inject } from "@nestjs/common";
-import { isNotNull, sql, and, eq, sum, asc, gte, lte } from "drizzle-orm";
+import { isNotNull, sql, and, eq, sum, asc, gte, lte, or } from "drizzle-orm";
 import { DataRetrievalException } from "src/exception";
 import {
   CapitalProjectCategory,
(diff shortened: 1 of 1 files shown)


TangoYankee (approved):
:rocket: :moon:

PR #439, opened 2025-03-27
Issue 435 Sort Agencies by Alphabetically by Initials

### LINKED ISSUE #435: Sort agencies by alphabetically by initials
# Description
Return the agencies sorted by their initials

## Acceptance criteria
- [x] Update the OpenAPI documentation to indicate `agencies` `order` is `initials`
- [x] Update the query in the repository to `orderBy` `initials`
- [x] Update the `service` to add the `order` to the results
- [x] Update tests to check for `order`

horatiorosa:
#### Description 
- update open api docs to add order to Agency + regen resources
 - Added sorting by order to the agency repository, service, and tests

#### Tickets
Closes #435

### DIFF
--- a/openapi/paths/agencies.yaml
+++ b/openapi/paths/agencies.yaml
@@ -5,7 +5,8 @@ get:
     - Agencies
   responses:
     '200':
-      description: An object containing all agencies.
+      description: >
+        An object containing all agencies sorted alphabetically by the agency initials.
       content:
         application/json:
           schema:
@@ -13,10 +14,18 @@ get:
             properties:
               agencies:
                 type: array
+                description: >
+                  An list of agencies sorted alphabetically by the agency initials.
                 items:
                   $ref: ../components/schemas/Agency.yaml
+              order:
+                type: string
+                description: >-
+                  The criteria used to sort the results using the agency initials in ascending order.
+                example: initials
             required:
               - agencies
+              - order
     '400':
       $ref: ../components/responses/BadRequest.yaml
     '500':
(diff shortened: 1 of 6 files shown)


openapi/components/schemas/Agency.yaml
```diff
@@ -8,6 +8,12 @@ properties:
     type: string
     description: The full name of the agency.
     example: Department of Transportation
+  order:
+    type: string
+    description: >-
+      The criteria used to sort the results using the agency initials in ascending order.
+    example: DCP
```
TangoYankee, line 15: The order informs the field on which the list is sorted, rather than a specific entry in the field.

So, it says "initials" instead of "DCP"

```suggestion
    example: initials
```

openapi/components/schemas/Agency.yaml
```diff
@@ -8,6 +8,12 @@ properties:
     type: string
     description: The full name of the agency.
     example: Department of Transportation
+  order:
```
TangoYankee, line 11: Similar to [agencyBudgets](https://github.com/NYCPlanning/ae-zoning-api/blob/3753599daea0c558fab003d71b38330ee1f529bd/openapi/paths/agency-budgets.yaml), `order` is a property on the path, rather than the schema (The schema being concerned only with fields that directly belong to "agencies"). I think  [`openapi/paths/agencies.yaml`](https://github.com/NYCPlanning/ae-zoning-api/blob/3753599daea0c558fab003d71b38330ee1f529bd/openapi/paths/agencies.yaml) would be a better spot

TangoYankee (changes requested):

TangoYankee (approved):

PR #441, opened 2025-04-03
Bugfix: commitment totals encoded as string

### LINKED ISSUE #440: Bug: Project totals incorrectly encoded in MVT as string
> the data in the map tiles needs to be an integer, not a string
>  

 _Originally posted by @TangoYankee in [#159](https://github.com/NYCPlanning/ae-cp-map/issues/159#issuecomment-2773376366)_

TangoYankee:
Coerce commitment totals from numeric to double precision type to encode them in mvts as numbers instead of strings

closes #440


dhochbaum-dcp (approved):
Ahh, I see, that's why just adding `.mapWith(Number)` didn't work for me.

PR #445, opened 2025-04-07
433/cache validations passing

### LINKED ISSUE #433: Cache validation results
# Description

We have several validations that rely on making O(n) requests to the database to check for the existance of an entry. The results are a lightweight boolean. Because the operation is relatively expensive, is repeated on each request, and has a lightwieght response, they are perfect candidates for caching.

## Acceptance criteria
- [ ] Cache the outcome of database calls that check for the presence of a value

TangoYankee:
Cache validation queries

closes #433

### DIFF
--- a/src/agency-budget/agency-budget.repository.ts
+++ b/src/agency-budget/agency-budget.repository.ts
@@ -6,11 +6,14 @@ import {
   FindManyRepo,
 } from "./agency-budget.repository.schema";
 import { agencyBudget } from "src/schema";
+import { Cache } from "cache-manager";
+import { CACHE_MANAGER } from "@nestjs/cache-manager";
 
 export class AgencyBudgetRepository {
   constructor(
     @Inject(DB)
     private readonly db: DbType,
+    @Inject(CACHE_MANAGER) private readonly cacheManager: Cache,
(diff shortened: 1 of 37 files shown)


src/agency-budget/agency-budget.repository.ts
```diff
@@ -21,9 +24,21 @@ export class AgencyBudgetRepository {
...
 
-  async checkByCode(code: string): Promise<CheckByCodeRepo | undefined> {
+  async checkByCode(code: string): Promise<CheckByCodeRepo> {
+    const key = JSON.stringify({
+      code,
+      domain: "agencyBudget",
+      function: "checkByCode",
+    });
+
+    const cachedValue: boolean | null = await this.cacheManager.get(key);
```
TylerMatteo, line 34: Just to sanity check my understanding of this code, when you do `cacheManager.get`, that function checks to see if a cached value is stale and returns null if it is? Or does that happen somewhere else?

horatiorosa: Correct, that is what happens:
[link to docs](https://www.npmjs.com/package/cache-manager#get)

Issue #467, opened 2025-06-25
Provide more details in data retrieval exceptions


TangoYankee:
## Acceptance criteria
- [x] Update `DataRetrievalException` to require a string with a brief detail for that error
  - Add detail to existing "base" data retrieval error message 
- [x] Update each instance of `DataRetrievalException` to have a brief detail
- [x] Update test cases to assert detail exists in message

PR #473, opened 2025-07-14
More details to data retrival exception

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/467


TangoYankee (changes requested):
The changes to tests make sense. We mock the repos so it makes sense that we're also mocking the thrown exceptions.

Great work, just a couple of pedantic changes.

And .. something... something... clean commit history

TangoYankee (approved):

Issue #465, opened 2025-06-09
Create CBBR tables


pratishta:
### Acceptance criteria
- [ ] Define the tables `community-board-budget-request`, `agency-response` and `community-board-budget-request-geom` in the cbbr schema

PR #492, opened 2025-09-03
Update schema and run generate

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/465


src/schema/community-board-budget-request.ts
```diff
@@ -49,3 +62,62 @@ export const cbbrOptionCascade = pgTable(
...
+  },
+);
+export const communityBoardBudgetRequest = pgTable(
+  "community_board_budget_request",
+  {
+    id: text("id").notNull().primaryKey(),
+    trackingCode: text("tracking_code").notNull(),
+    boroughId: char("borough_id", { length: 1 })
+      .notNull()
+      .references(() => borough.id),
```
TangoYankee, line 80: I think it would be clearer if there were only one reference for the borough_id. I would lean toward the composite definition on line 107 because it's more powerful

 
```suggestion
```

src/schema/community-board-budget-request.ts
```diff
@@ -49,3 +62,62 @@ export const cbbrOptionCascade = pgTable(
...
+      "community_board_budget_request_request_type_options",
+      sql`${table.requestType} IN (
+        'Capital',
+        'Expense'
+      )`,
+    ),
+    index().using("GIST", table.liFtMPnt),
+    index().using("GIST", table.liFtMPoly),
+    index().using("GIST", table.mercatorFillMPoly),
+    index().using("GIST", table.mercatorFillMPnt),
```
TangoYankee, line 121: What do you think of adding a spatial index for the mercator label column?

src/schema/community-board-budget-request.ts
```diff
@@ -49,3 +62,62 @@ export const cbbrOptionCascade = pgTable(
...
+  "cbbr_agency_category_response",
+  {
+    id: smallint("id").generatedByDefaultAsIdentity().primaryKey(),
+    description: text("description").notNull(),
+  },
+);
+export const communityBoardBudgetRequest = pgTable(
+  "community_board_budget_request",
+  {
+    id: text("id").notNull().primaryKey(),
```
TangoYankee, line 76: I just want to make sure things are clear for everyone between the use of "id" in the api and "id" in the database. In the api "id" refers to "tracking code". In the database "id" refers to "id"

TangoYankee (changes requested):
A couple small items. Looking good overall

src/schema/community-board-budget-request.ts
```diff
@@ -49,3 +62,62 @@ export const cbbrOptionCascade = pgTable(
...
+      foreignColumns: [communityDistrict.boroughId, communityDistrict.id],
+    }),
+    check(
+      "community_board_budget_request_request_type_options",
+      sql`${table.requestType} IN (
+        'Capital',
+        'Expense'
+      )`,
+    ),
+    index().using("GIST", table.liFtMPnt),
```
TangoYankee, line 118: The API will be looking up budget requests based on the tracking code. It might be worth considering adding an index to that column, as well

src/schema/community-board-budget-request.ts
pratishta: As per our discussion with David, I updated id to refer to the tracking_code and internal_id to refer to the unique_id.

TangoYankee (dismissed):
Looks good :rocket: :moon: 

I left a comment about indexing the "id". I believe it's now superfluous but harmless. I have no strong opinions on whether to keep or discard the code that manually creates the index

TangoYankee (approved):

PR #495, opened 2025-09-11
487/policy area test data

TangoYankee:
## Mocks
- substituting hand-written test data for auto-generated mock data
  - this enforces the shape and relationships of the data, without coupling it to the actual content
- construct mock cbbr options with data from other mock tables
- refactor mock findPolicyAreas
  - Update types from "null" to "undefined", to match the real function
  - Update the logical check to "escape when a policy area fails to meet a criteria"

## Missing agency message
- Update the "agency initials" error message to specificy that they're missing, which is the specific way they're invalid

## Missing need group id
- Add a check for missing need group ids.
  - Checks for agency initials and need group ids weren't part of the acceptance criteria of this ticket. I figured we could make them a follow-up ticket. But if the agency check and the cachemanager are already included and, then I thought I'd sneak in need group id, as well

## Additional test cases
- Add a service test for missing agency initials
- Add service and e2e tests from missing need group ids


dhochbaum-dcp:
Added to branch.

Issue #489, opened 2025-08-27
Add /community-board-budget-requests/agencies endpoint


dhochbaum-dcp:
### `/community-board-budget-requests/agencies`
- operationId: findCommunityBoardBudgetRequestAgencies
- Response codes
  - 200
  - 400
  - 500 
- query parameters
  - cbbrPolicyAreaId, integer, optional
  - cbbrNeedGroupId, integer, optional
- response body
  - initials, string, required
  - name, string, required

## Acceptance Criteria
Update the OpenAPI docs:
- [ ] Add /community-board-budget-requests/agencies path

Create the endpoint:
Endpoint should be `/community-board-budget-requests/agencies`.
- [ ] Endpoint should be reflected in OpenAPI spec (regenerate files as needed)
- [ ] Add tests for new endpoint

PR #497, opened 2025-09-17
489 Add /community-board-budget-requests/agencies endpoint

dhochbaum-dcp:
### `/community-board-budget-requests/agencies`
- operationId: findCommunityBoardBudgetRequestAgencies
- Response codes
  - 200
  - 400
  - 500 
- query parameters
  - cbbrPolicyAreaId, integer, optional
  - cbbrNeedGroupId, integer, optional
- response body
  - initials, string, required
  - name, string, required

## Acceptance Criteria
Update the OpenAPI docs:
- [x] Add /community-board-budget-requests/agencies path

Create the endpoint:
Endpoint should be `/community-board-budget-requests/agencies`.
- [x] Endpoint should be reflected in OpenAPI spec (regenerate files as needed)
- [x] Add tests for new endpoint

Completes #489

### DIFF
--- a/src/community-board-budget-request/community-board-budget-request.repository.ts
+++ b/src/community-board-budget-request/community-board-budget-request.repository.ts
@@ -6,10 +6,17 @@ import {
   CheckPolicyAreaByIdRepo,
   FindNeedGroupsRepo,
   FindPolicyAreasRepo,
+  FindAgenciesRepo,
 } from "./community-board-budget-request.repository.schema";
-import { cbbrNeedGroup, cbbrPolicyArea, cbbrOptionCascade } from "src/schema";
-import { eq, and, sql } from "drizzle-orm";
 import {
+  agency,
+  cbbrNeedGroup,
+  cbbrPolicyArea,
+  cbbrOptionCascade,
+} from "src/schema";
+import { eq, and, sql, isNotNull } from "drizzle-orm";
+import {
+  FindCommunityBoardBudgetRequestAgenciesQueryParams,
   FindCommunityBoardBudgetRequestNeedGroupsQueryParams,
   FindCommunityBoardBudgetRequestPolicyAreasQueryParams,
 } from "src/gen";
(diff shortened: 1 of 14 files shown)


src/community-board-budget-request/community-board-budget-request.repository.ts
```diff
@@ -85,6 +92,38 @@ export class CommunityBoardBudgetRequestRepository {
...
+          name: agency.name,
+        })
+        .from(agency)
+        .leftJoin(
+          cbbrOptionCascade,
+          eq(agency.initials, cbbrOptionCascade.agencyInitials),
+        )
+        .where(
+          and(
+            isNotNull(cbbrOptionCascade),
```
TangoYankee, line 112: As far as I can tell, `IS NOT NULL` is intended to check a specific column. For this case of checking the whole table, we are getting the expected results. However, I can't find an explanation for how it's working. My guess is that it returns `true` when every column in the table is `NULL`?

I'm concerned about future changes to the data causing the query to behave in unexpected ways. I think the suggestions below would be more resilient to changes.

Either:
 1) Checking the "agency initials" column
```typescript
    isNotNull(cbbrOptionCascade.agencyInitials),
```

Or 
2) replacing the "left join & where" with "right join"
```typescript
   .selectDistinct({
          initials: agency.initials,
          name: agency.name,
        })
        .from(agency)
        .rightJoin(
          cbbrOptionCascade,
          eq(agency.initials, cbbrOptionCascade.agencyInitials),
        )
         .where(
          and(
            cbbrNeedGroupId !== undefined
              ? eq(cbbrOptionCascade.needGroupId, cbbrNeedGroupId)
              : undefined,
            cbbrPolicyAreaId !== undefined
              ? eq(cbbrOptionCascade.policyAreaId, cbbrPolicyAreaId)
              : undefined,
          ),
        )
        .orderBy(agency.name);
```

TangoYankee (changes requested):

TangoYankee (approved):

PR #586, opened 2026-03-16
565 Filter capital projects by multiple city council districts

### LINKED ISSUE #565: Filter capital projects by multiple city council districts
# Description

Add a query parameter to filter capital projects by city council districts

## Acceptance criteria
- [ ] Add a `cityCouncilDistrictIds` query parameter to the 'findCapitalProjects' operation
- [ ] Validate the city council district ids parameter:
   - is a list of ids
   - has a length between 1 and 10, inclusive
   - each id matches the existing structure of a city council id
   - every id exists in the database
- [ ] When searching multiple city council districts, return projects that are in any of the provided city council districts
- [ ] When combined with another geographic filter, return only projects that are in both geographies

#### Validating all requested city council districts exist

```sql
WITH city_council_district_filter AS (
	SELECT UNNEST(ARRAY['3', '2', '79']) AS id
)
SELECT city_council_district_filter.id FROM city_council_district_filter
LEFT JOIN city_council_district
  ON city_council_district_filter.id = city_council_district.id
WHERE city_council_district.id IS NULL;
```

**Update:** Drizzle doesn't have an idiomatic way to implement the above validation. When attempting to use this approach, I found it could only be achieved by placing the whole query in the "sql magic operator". There also isn't much benefit to listing all the unfound ids as our error detection does a simple true/false check. It would be simpler to instead deduplicate the ids and then checked with the number of requested ids and the number of found ids matched. 

#### Finding capital projects that intersect with any of the city council districts

(This shows a simplified version that looks only at the multipolygons of capital projects. The full version should also look at the multipoints)

```sql
SELECT 
	capital_project.managing_code, 
	capital_project.id,
	capital_project.li_ft_m_poly
FROM
capital_project
LEFT JOIN city_council_district
	ON ST_INTERSECTS(city_council_district.li_ft, capital_project.li_ft_m_poly)
WHERE capital_project.li_ft_m_poly IS NOT NULL
	AND city_council_district.id = ANY(ARRAY['1', '2', '3', '4']);
```

dhochbaum-dcp:
Adds filtering by multiple city council districts, leaves cityCouncilDistrictId parameter as-is.

Closes #565

### DIFF
--- a/src/city-council-district/city-council-district.repository.ts
+++ b/src/city-council-district/city-council-district.repository.ts
@@ -3,6 +3,7 @@ import { DB, DbType } from "src/global/providers/db.provider";
 import { DataRetrievalException } from "src/exception";
 import {
   CheckByIdRepo,
+  CheckByIdsRepo,
   FindManyRepo,
   FindTilesRepo,
   FindGeoJsonByIdRepo,
@@ -23,7 +24,7 @@ import {
   cityCouncilDistrict,
   communityBoardBudgetRequest,
 } from "src/schema";
-import { eq, sql, isNotNull, and, or } from "drizzle-orm";
+import { eq, sql, isNotNull, and, or, inArray } from "drizzle-orm";
 import { Cache } from "cache-manager";
 import { CACHE_MANAGER } from "@nestjs/cache-manager";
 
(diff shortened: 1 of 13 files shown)


src/city-council-district/city-council-district.repository.ts
```diff
@@ -67,6 +68,35 @@ export class CityCouncilDistrictRepository {
...
+      const ccds = await this.db
+        .select({ count: sql<number>`count(${cityCouncilDistrict.id})::int` })
+        .from(cityCouncilDistrict)
+        .where(inArray(cityCouncilDistrict.id, ids));
+      const value = ccds[0].count == ids.length;
+      this.cacheManager.set(key, value);
+      return value;
+    } catch {
+      throw new DataRetrievalException(
+        "cannot check for city council districts based on multiple ids",
```
TylerMatteo, line 95: So this gets thrown in the case where a user sent an ID that wasn't found in the database? And then that leads to this throw on line 181 of `capital-project.service.ts`?
```
    if (checkedList.some((result) => result === false))
      throw new InvalidRequestParameterException(
        "could not check one or more of the parameters",
      );
```

Unless I'm missing something (which is very possible), the `"cannot check for city council districts based on multiple ids"` exception never actually gets logged or bubbled up. Is that accurate?

dhochbaum-dcp: It won't be logged/bubbled up if the check successfully returns false.  It will be returned in the event of an error in the process of checking, such as if the database is down.

TylerMatteo (approved):

PR #611, opened 2026-05-05
Added Borough Id to Census Tract Table

### LINKED ISSUE #610: Update census tract schema to include boroughs
### Acceptance criteria
- [x] Add borough as a column with type char and length of 1 to the census tract table. It should reference the `borough` table. Generate the migration and run it
- [x] Have boroughs be part of the composite key along with label and year. Generate the migration and run it.

Note: We will generate two migrations because adding the column and using it as part of the composite key can cause 'this column doesn't yet exist' issues.

horatiorosa:
#### Description
Added `boroughId` to `census_tract` table for use in the composite primary key

#### Ticket
Closes #610


pratishta (approved):
Nice

PR #614, opened 2026-05-07
567 Filter community board budget requests by multiple city council districts

### LINKED ISSUE #567: Filter community board budget requests by multiple city council districts
# Description

Add a query parameter to filter community board budget requests by city council districts.

## Acceptance criteria
- [ ] Add a `cityCouncilDistrictIds` query parameter to the 'findCommunityBoardBudgetRequests' operation
- [ ] Validate the city council district ids parameter:
   - is a list of ids
   - has a length between 1 and 10, inclusive
   - each id matches the existing structure of a city council district id
   - every id exists in the database
- [ ] When searching multiple city council districts, return projects that are in any of the provided city council districts
- [ ] When combined with another geographic filter, return only projects that are in both geographies

### Dependent on
#565 for query parameter and code to check for multiple city council districts

#### Validating all requested city council districts exist

```sql
WITH city_council_district_filter AS (
	SELECT UNNEST(ARRAY['3', '2', '79']) AS id
)
SELECT city_council_district_filter.id FROM city_council_district_filter
LEFT JOIN city_council_district
  ON city_council_district_filter.id = city_council_district.id
WHERE city_council_district.id IS NULL;
```

**Update:** Drizzle doesn't have an idiomatic way to implement the above validation. When attempting to use this approach, I found it could only be achieved by placing the whole query in the "sql magic operator". There also isn't much benefit to listing all the unfound ids as our error detection does a simple true/false check. It would be simpler to instead deduplicate the ids and then checked with the number of requested ids and the number of found ids matched. 
#### Finding community board budget requests that intersect with any of the city council districts

(This shows a simplified version that looks only at the multipolygons of community board budget requests. The full version should also look at the multipoints)

```sql
SELECT 
	community_board_budget_request.id, 
	community_board_budget_request.li_ft_m_poly
FROM
community_board_budget_request
LEFT JOIN city_council_district
	ON ST_INTERSECTS(city_council_district.li_ft, community_board_budget_request.li_ft_m_poly)
WHERE community_board_budget_request.li_ft_m_poly IS NOT NULL
	AND city_council_district.id = ANY(ARRAY['1', '2', '3', '4']);
```

dhochbaum-dcp:
Added CBBR filtering by cityCouncilDistrictIds.

This PR includes both the JSON and CSV endpoints.

Closes #567


src/community-board-budget-request/community-board-budget-request.repository.ts
```diff
@@ -317,18 +319,17 @@ export class CommunityBoardBudgetRequestRepository {
...
-          cbbrPolicyAreaId: communityBoardBudgetRequest.policyArea,
-          title: communityBoardBudgetRequest.title,
-          communityBoardId: sql<string>`${borough.abbr} || ${communityBoardBudgetRequest.communityDistrictId}`,
-          isMapped: sql<boolean>`${or(
-            isNotNull(communityBoardBudgetRequest.liFtMPnt),
-            isNotNull(communityBoardBudgetRequest.liFtMPoly),
-          )}`,
-          isContinuedSupport: communityBoardBudgetRequest.isContinuedSupport,
-        })
+      return await this.db[geom === null ? "selectDistinct" : "select"]({
```
dhochbaum-dcp, line 325: Because a CBBR can intersect with multiple boroughs or CCDs, the select statement was returning duplicate CBBRs, requiring the switch from using `select` to `selectDistinct`.  Unfortunately, there is an issue with drizzle-orm with `selectDistinct` used with `orderBy` - it was hard to pin down exactly but the query we use to sort by distance in the `orderBy` cannot be used with `selectDistinct`.  My solution was if we are filtering by radius, to use `select` with the sort by distance in the `orderBy`; otherwise we use `selectDistinct` and sort by id.

TylerMatteo: Is the Drizzle issue something you've seen reported in their Issues by others or just something you noticed here?

dhochbaum-dcp: I recall seeing it in the AI summary of one of my searches, I'm having trouble finding info about it now.  I think it may be related to Postgres requiring that anything in the orderBy also be in the select.

TylerMatteo: I'm less concerned about this given that we don't currently let users filter by a CCD and a radius at the same time, so this isn't really losing any functionality. Please just leave a comment explaining the context for the change and should be good to go.

TylerMatteo (dismissed):

TylerMatteo (approved):

PR #616, opened 2026-05-11
Add facility operator

### LINKED ISSUE #602: Create facility_operator table
### Acceptance criteria
- [ ] Define the table `facility_operater` in `facility.ts` schema file

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/602

Went with using a generated uid because there are facility operators with a null abbreviation but an existing name (and vice versa)

### DIFF
--- a/db/migration/meta/_journal.json
+++ b/db/migration/meta/_journal.json
@@ -246,6 +246,13 @@
       "when": 1777990120474,
       "tag": "0034_glorious_the_captain",
       "breakpoints": true
+    },
+    {
+      "idx": 35,
+      "version": "7",
+      "when": 1778613108863,
+      "tag": "0035_dear_serpent_society",
+      "breakpoints": true
     }
   ]
 }
--- a/src/schema/facility-operator.ts
+++ b/src/schema/facility-operator.ts
@@ -1,0 +1,38 @@
+import { pgTable, text, uuid, check } from "drizzle-orm/pg-core";
+import z from "zod";
+import { sql } from "drizzle-orm";
+
+export const facilityOperator = pgTable(
+  "facility_operator",
+  {
+    id: uuid("id").primaryKey(),
+    abbreviation: text("abbreviation"),
+    name: text("name"),
+    type: text("type"),
+  },
+  (table) => [
+    check(
+      "facility_operator_type_options",
+      sql`${table.type} IN (
+      'Public',
+      'Non-public'
+    )`,
+    ),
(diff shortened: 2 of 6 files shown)


db/migration/meta/_journal.json
```diff
@@ -246,6 +246,13 @@
       "when": 1777990120474,
       "tag": "0034_glorious_the_captain",
       "breakpoints": true
+    },
+    {
+      "idx": 35,
+      "version": "7",
+      "when": 1778601984650,
+      "tag": "0035_futuristic_dragon_man",
```
horatiorosa: i love these migration names

src/schema/facility-operator.ts
```diff
@@ -0,0 +1,38 @@
...
+  "facility_operator",
+  {
+    id: uuid("id").primaryKey(),
+    abbreviation: text("abbreviation"),
+    name: text("name"),
+    type: text("type"),
+  },
+  (table) => [
+    check(
+      "facility_operator_$$type_options",
```
horatiorosa: Do the `$$` serve a specific function? None of the other check constraints use this naming pattern.

src/schema/facility-operator.ts
```diff
@@ -0,0 +1,38 @@
...
+      sql`${table.type} IN (
+      'Public',
+      'Non-public'
+    )`,
+    ),
+  ],
+);
+
+export const facilityOperatorTypeEntitySchema = z.enum([
```
horatiorosa, line 24: it wasn't possible to add facility operator to the schema to the open api docs?

pratishta: This brings up a good point of when it is a good idea to have a component schema defined and when to have it just be a type and check constraint. We added CapitalProjectCategory (and soon AgencyOversightLevelCategory) to Open API, presumably because we thought they would be reused throughout the API docs. So having a schema would make that easier. 

For this FacilityOperatorType (as well as CBBR request type between Capital vs Expense), I don't imagine we would use these defined schema objects in other API definitions. 

That said, we still only use CapitalProjectCategory in one place (same with AgencyOversightLevelCategory) so we should come up with a standard or a documented reason for when we add to Open API. 

@TylerMatteo

src/schema/facility-operator.ts
pratishta: That was a mistake

TylerMatteo (approved):

PR #628, opened 2026-06-08
Facilities schema

### LINKED ISSUE #627: Add facilities table schema
wip

pratishta:
Closes https://github.com/NYCPlanning/ae-zoning-api/issues/627

#### Testing:

- Run `npm run drizzle:migrate` and it should create the facility table with the appropriate constraints and relations

### DIFF
--- a/src/schema/facility.ts
+++ b/src/schema/facility.ts
@@ -1,5 +1,21 @@
-import { pgTable, smallint, text } from "drizzle-orm/pg-core";
+import {
+  char,
+  check,
+  index,
+  pgTable,
+  smallint,
+  text,
+  uuid,
+} from "drizzle-orm/pg-core";
 import z from "zod";
+import {
+  facilityOperator,
+  facilityOperatorEntitySchema,
+} from "./facility-operator";
+import { agency, agencyEntitySchema } from "./agency";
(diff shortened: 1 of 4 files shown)


src/schema/facility.ts
```diff
@@ -72,3 +88,65 @@ export const facilityTypeEntitySchema = z.object({
...
+    name: text("name"),
+    address: text("address"),
+    addressNumber: text("address_number"),
+    streetName: text("street_name"),
+    city: text("city"),
+    zipCode: text("zip_code"),
+    facilityTypeId: smallint("facility_type_id")
+      .notNull()
+      .references(() => facilityType.id),
+    serviceArea: text("service_area"),
```
horatiorosa: should `serviceArea` be `notNull()` since we have the check constraint expecting only local or regional?

dhochbaum-dcp (dismissed):

src/schema/facility.ts
pratishta: Good point, I assumed that the check constraint would take care of that but it seems that unless we explicitly set notNull(), it would still accept null values. Updated.

dhochbaum-dcp (approved):

PR #630, opened 2026-06-11
626 Create FacDB OpenAPI Schema

### LINKED ISSUE #626: Create FacDB OpenAPI Schema
As described in https://github.com/NYCPlanning/ae-zoning-api/discussions/621#discussioncomment-17020188

dhochbaum-dcp:
This updates the OpenAPI schema for all 6 planned endpoints.

As described here: https://github.com/NYCPlanning/ae-zoning-api/discussions/621#discussioncomment-17020188

Closes #626


openapi/components/parameters/bblQueryParam.yaml
```diff
@@ -0,0 +1,10 @@
+name: bbl
+required: false
+in: query
+schema:
+  type: string
+  pattern: ^([0-9]{10})$
+  example: '1000477501'
+description: >-
+  The ten character code compromised of a one character borough, five character
```
horatiorosa: spelling : I assume this should be `comprised`

dhochbaum-dcp: Good catch, thanks

openapi/components/schemas/Facility.yaml
```diff
@@ -0,0 +1,75 @@
...
+    type: string
+    description: The seven digit building id of the facility.
+    example: '1045270'
+    nullable: true
+  facilityJurisdiction:
+    $ref: ./OversightLevelCategory.yaml
+  facilityOperatorType:
+    $ref: ./FacilityOperatorType.yaml
+  categoryId:
+    type: number
```
horatiorosa: I am wondering if all these id's (categoryId, categoryGroupId, categorySubgroupId) should be ` type: string` rather than `type: number`. The params (facilityCategoryIdsQueryParam.yaml, facilityCategoryGroupIdsQueryParam.yaml, and facilityCategorySubgroupIdsQueryParam.yaml) all use `type: string`. The mismatch might cause wonky side effects

openapi/components/schemas/FacilityPage.yaml
```diff
@@ -0,0 +1,35 @@
...
+              description: The id for the facility or program.
+              example: 1435f9fd9f0787a5038b15512bb9684f
+            name:
+              type: string
+              description: The name of the facility or program.
+              example: 'QUEENSBORO OVAL'
+            oversightAgency:
+              type: string
+              description: The name of the agency overseeing the facility or program.
+              example: 'NYC Department of Parks and Recreation'
```
horatiorosa, line 22: `oversightAgency` and `name` should be `nullable: true` bc Facility yaml has the values as [nullable](https://github.com/NYCPlanning/ae-zoning-api/pull/630/changes#diff-4f41ee1b46791212e94da1c072a115fd466dfc17f829f990073f33d7b75fdfacR16)

openapi/components/schemas/FacilityCategorySubgroup.yaml
```diff
@@ -0,0 +1,19 @@
...
+    description: The id for the subgroup.
+    example: '1'
+  name:
+    type: string
+    description: The name of the subgroup.
+    example: Public K-12 Schools
+  description:
+    type: string
+    description: The description of the subgroup.
+    example: Public K-12 Schools
```
horatiorosa, line 15: the description in FacCatGroup is nullable but this one is not.  Just wanna make sure adding nullable here wasn't overlooked

dhochbaum-dcp: Good thinking, but this is correct, there is a blank group description, but none of the subgroup ones appeared to be blank

horatiorosa: none are blank, for now. we never know what the data will be like down the road.

openapi/components/schemas/Facility.yaml
dhochbaum-dcp: Good call, I've made that update.

openapi/components/schemas/FacilityPage.yaml
dhochbaum-dcp: Good catch.  I've also added `categoryId` based upon updates to the designs.

openapi/components/schemas/FacilityCategorySubgroup.yaml
dhochbaum-dcp: This field is just used for tooltips, we copied them from the old Capital Planning Explorer.  They are not updated regularly, I don't anticipate anyone deciding to remove an existing tooltip or add a new subgroup without a tooltip.

openapi/components/schemas/FacilityPage.yaml
```diff
@@ -0,0 +1,40 @@
...
+              description: The name of the facility or program.
+              example: 'QUEENSBORO OVAL'
+              nullable: true
+            oversightAgency:
+              type: string
+              description: The name of the agency overseeing the facility or program.
+              example: 'NYC Department of Parks and Recreation'
+              nullable: true
+            categoryId:
+              type: number
```
horatiorosa: I missed this one,  another mismatch, `type: number` here but `type: string` in [Facility.yaml](https://github.com/NYCPlanning/ae-zoning-api/pull/630/changes#diff-4f41ee1b46791212e94da1c072a115fd466dfc17f829f990073f33d7b75fdfacR37)

dhochbaum-dcp: Nope, that error is all me, it's a freshly introduced one.

horatiorosa (dismissed):
LGTM but it's a good idea to have @pratishta  take a look too before merging.

src/schema/facility.ts
```diff
@@ -45,7 +46,7 @@ export const facilityGroup = pgTable("facility_group", {
 export const facilityGroupEntitySchema = z.object({
   id: z.number().int(),
   name: z.string(),
-  description: z.string(),
+  description: z.string().nullable(),
```
pratishta, line 49: If this is because we don't have a description for 'Adult Services' we should ask DE if that's a bug / we can fill it in. Let's not specify that it's nullable (since it is nullable implicitly now).

openapi/components/schemas/Facility.yaml
```diff
@@ -0,0 +1,78 @@
...
+    pattern: ^([0-9]{1,3})$
+    example: '1'
+  operatorName:
+    type: string
+    description: The name of the entity operating the facility or program.
+    example: 'NYC Department of Parks and Recreation'
+    nullable: true
+  dataSource:
+    $ref: ./DataSource.yaml
+  alsoAtLocation:
```
pratishta, line 58: Is there a reason why we would want to have also at location to be part of `/facilities/{facilityId}`? Instead of calling it separately via `/facilities?bbl=facilityId.bbl` to populate that section of the facility detail panel?

openapi/components/schemas/FacilityGeoJson.yaml
```diff
@@ -0,0 +1,21 @@
...
+    description: The id for the facility.
+    example: 1435f9fd9f0787a5038b15512bb9684f
+  type:
+    type: string
+    enum:
+      - Feature
+  geometry:
+    oneOf:
+      - $ref: ./MultiPoint.yaml
+      - $ref: ./MultiPolygon.yaml
```
pratishta: We don't have polygon data for facilities so we can get rid of this

openapi/paths/facilities_agencies.yaml
```diff
@@ -0,0 +1,24 @@
+get:
+  summary: Find facility agencies
```
pratishta, line 2: For some reason I'm unable to see the summary on the OpenAPI docs, is it showing up for you? 

Image

openapi/components/schemas/Facility.yaml
```diff
@@ -0,0 +1,78 @@
...
+    type: string
+    description: The category subgroup id of the facility
+    pattern: ^([0-9]{1,3})$
+    example: '1'
+  operatorName:
+    type: string
+    description: The name of the entity operating the facility or program.
+    example: 'NYC Department of Parks and Recreation'
+    nullable: true
+  dataSource:
```
pratishta, line 56: Why are we returning the entire data source entity and not just dataSourceSchema (which is its id)?

openapi/components/schemas/Facility.yaml
```diff
@@ -0,0 +1,78 @@
...
+  id:
+    type: string
+    description: The id for the facility or program.
+    example: 1435f9fd9f0787a5038b15512bb9684f
+  name:
+    type: string
+    description: The name of the facility or program.
+    example: 'QUEENSBORO OVAL'
+    nullable: true
+  oversightAgency:
```
pratishta: Should this be oversightAgencyInitials?

pratishta (changes requested):

openapi/components/schemas/Facility.yaml
dhochbaum-dcp: Yes:
 - Whenever we are using the `/facilities/{facilityId}` endpoint, we will always be using the alsoAtLocation data, so it doesn't make sense to make the request, wait to parse the response, and make another request
 - We only need the `id` and `name` fields, so the `/facilities` endpoint would return a lot of unnecessary information
 - The alsoAtLocation filters by `bin OR bbl`, rather than `bin AND bbl`, which would mean making a lot of extra design decisions

openapi/components/schemas/FacilityGeoJson.yaml
dhochbaum-dcp: Good catch.

openapi/paths/facilities_agencies.yaml
dhochbaum-dcp: We don't show that summary in the docs, we only show the summaries for the tags.

openapi/components/schemas/Facility.yaml
dhochbaum-dcp: The same reasons as for alsoAtLocation

openapi/components/schemas/Facility.yaml
dhochbaum-dcp: Yes

src/schema/facility.ts
dhochbaum-dcp: DE did not provide that data, I pulled it from the old site, so it's not a bug.  Let's just specify that it's nullable.

pratishta (approved):

Issue #634, opened 2026-06-11
CPP/FacDB - Add /facilities/{z}/{x}/{y}.pbf endpoint


dhochbaum-dcp:
Should include:

- Implementation of the endpoint as described: https://github.com/NYCPlanning/ae-zoning-api/discussions/621#discussioncomment-17020188

PR #647, opened 2026-06-22
634 - CPP/FacDB - Add /facilities/{z}/{x}/{y}.pbf endpoint

dhochbaum-dcp:
Completes #634


pratishta (approved):
lgtm, able to hit endpoint with default params and get a data-ful pbf file.

PR #659, opened 2026-06-29
658 Add URL to dataSource in /facility/{facilityId} Endpoint

### LINKED ISSUE #658: CPP/FacDB - Add URL to dataSource in /facility/{facilityId} Endpoint

dhochbaum-dcp:
Closes #658

### DIFF
--- a/openapi/components/schemas/DataSource.yaml
+++ b/openapi/components/schemas/DataSource.yaml
@@ -18,8 +18,13 @@ properties:
     type: string
     description: The retrieval date of the version of the dataset.
     example: 2026-04-17
+  url:
+    type: string
+    description: The URL at which users can find the dataset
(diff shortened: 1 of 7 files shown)


src/schema/data-source.ts
```diff
@@ -6,13 +6,15 @@ export const dataSource = pgTable("data_source", {
   datasetName: text("dataset_name"),
   version: text("version").notNull(),
   retrieveDate: date("retrieve_date").notNull(),
+  url: text("url"),
```
pratishta, line 9: We already updated the schema in a previous PR so we don't need the changes to this file.

pratishta (changes requested):

src/schema/data-source.ts
dhochbaum-dcp: Merged those commits, so this file is no longer being updated.

pratishta (approved):

Issue #650, opened 2026-06-25
CPP/FacDB - Add short_name column to facility domain schema


pratishta:
Schema changes for https://github.com/NYCPlanning/ae-data-flow/issues/184

PR #662, opened 2026-07-02
635, 650 Fixes

dhochbaum-dcp:
This fixes a couple of issues:
#635 Orders the categories endpoint results by id
          Returns ids as integers instead of strings
#650 Updates the OpenAPI schema to add shortName to the categories endpoint


horatiorosa (approved):
LGTM

🚀

